PR #42, opened 2025-04-16
Upgrade protobuf for aarch64 compatibility

namloan:
There is no aarch64 binary in 3.15.0. Upgrading to 3.21.12 which is the latest version that has an aarch64 binary and keeps full backwards compatibility.

### DIFF
--- a/rpc/pom.xml
+++ b/rpc/pom.xml
@@ -49,9 +49,9 @@
                 <artifactId>protobuf-maven-plugin</artifactId>
                 <version>0.6.1</version>
                 <configuration>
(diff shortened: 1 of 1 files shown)


stejbac:
I incidentally upgraded `com.google.protobuf:protoc` to 3.25.6 and `io.grpc:protoc-gen-grpc-java` to 1.71.0 in an unrelated branch (https://github.com/stejbac/bisq-musig/tree/add-automated-rpc-tests) I was working on and hoping to submit soon as a PR. Do you know if that has an aarch64 target, or it is only specific versions? I've only been testing on Linux so far.

namloan:
All versions starting from 3.20.0 come with the aarch64 target. Canceling this PR in favor of yours then.

PR #43, opened 2025-04-16
Make `rpc` into a library crate and add automated tests

stejbac:
Add missing unit tests to `observable.rs` and rearrange the module hierarchy somewhat, placing the Prost-generated code under a `pb` parent module, along with the proto conversion logic moved from `server.rs`. Also make `rpc` into a library crate by adding `src/lib.rs` and making the binary `musigd` & `musig-cli` targets depend on it as an external crate, moving their main functions into the `src/bin` folder (as every Rust file there gets automatically picked up as a binary target of the same name by Cargo). This makes the project layout a bit more idiomatic.

Also add some preliminary integration tests, runnable by `cargo test` (along with the unit tests). These presently just do some rudimentary testing of the Wallet RPC service with an empty wallet, by spawning a transient gRPC server as a Tokio task and running the `musig-cli` binary client against it as a child process, checking its stdout/stderr output and exit code.

Finally, update the Java and Rust dependencies to their latest versions (as of a week or two ago) and do some additional minor code cleanup (such as newly triggered Clippy lints).

--

I'm presently still in the process of integrating Nigiri for more extensive integration testing of the wallet code against a _regtest_ bitcoind instance, and have not yet started proper integration tests for the protocol gRPC API either -- those will probably be done in a separate PR.


stejbac:
Currently, I'm encountering the following errors in the bitcoind logs when running Nigiri:

```
bitcoin  | 2025-04-16T15:26:57Z 
bitcoin  | 
bitcoin  | ************************
bitcoin  | EXCEPTION: St12system_error       
bitcoin  | Operation not permitted       
bitcoin  | bitcoin in AppInit()       
bitcoin  | 
bitcoin  | 
bitcoin  | 
bitcoin  | ************************
bitcoin  | EXCEPTION: St12system_error       
bitcoin  | Operation not permitted       
bitcoin  | bitcoin in AppInit()       
bitcoin  | 
bitcoin  | 2025-04-16T15:26:57Z Shutdown: In progress...
bitcoin  | 2025-04-16T15:26:57Z Shutdown: done
```

(Perhaps it's an AppArmor configuration issue? It may take a little while to get to the bottom of it and meanwhile I'll also check if I'm running into any similar Docker issues running the Bisq2 tests.)

namloan:
ACK

ChrisSon15:
> Currently, I'm encountering the following errors in the bitcoind logs when running Nigiri:
> 

One thing I obeserve while working with nigiri, is that after booting, I need to execute
````
nigiri stop
nigiri start
````
for it to properly startup. May be this is related to the same problem. I haven't figured out why I had to do this.

stejbac:
Stopping and starting Nigiri doesn't seem to help. I think it's just the bitcoind component that's broken -- Esplora displays on localhost:5000 but obviously gives an error, as it cannot connect to bitcoind. I'm guessing it's just something wrong with my Docker setup and the permissions it's granting / been granted -- I've never used it before and don't know much about it.

namloan:
@stejbac make sure you follow https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user. 

@ChrisSon15 I´ve tested this PR and works well, feel free to merge.

Let´s follow up on the Nigiri issues.

PR #48, opened 2025-05-03
Improve trade protocol client

HenrikJannsen:


HenrikJannsen:
@stejbac Could you review? 
I use the same class names and variable names on the Bisq 2 side, thus I tried to keep it in sync here.

HenrikJannsen:
Yes, have not followed up here. Not sure if @stejbac have had a look...

rpc/src/main/java/bisq/TradeProtocolClient.java
```diff
@@ -47,74 +47,79 @@ private static void testMusigService_twoParties(MusigGrpc.MusigBlockingStub stub
...
         System.out.println("Got reply: " + sellerDepositPsbt);
 
-        // Seller sends Message D to buyer.
+        // Seller sends Message D with sellerPartialSignaturesMessage to buyer.
 
+        // BUYER_AS_TAKER
         var buyerDepositPsbt = stub.signDepositTx(DepositTxSignatureRequest.newBuilder()
                 .setTradeId(buyerTradeId)
-                .setPeersPartialSignatures(sellerPartialSignatureMessage)
+                .setPeersPartialSignatures(sellerPartialSignaturesMessage) //TODO is clearSwapTxInputPartialSignature here needed as well?
```
stejbac, line 122: No, the seller's `swapTxInputPartialSignature` serves as the adaptor between the seller's private key share on the buyer payout and the final Swap Tx signature, so excluding it would prevent the buyer from verifying it (and thus that he can definitely claim his payout if the seller publishes the Swap Tx). Clearing it leads to an error response, thus preventing the trade from starting.

rpc/src/main/java/bisq/TradeProtocolClient.java
```diff
@@ -47,74 +47,79 @@ private static void testMusigService_twoParties(MusigGrpc.MusigBlockingStub stub
...
                 .setTradeId(buyerTradeId)
-                .setPeersNonceShares(sellerNonceShareMessage)
+                .setPeersNonceShares(sellerNonceSharesMessage)
                 .addAllReceivers(mockReceivers())
                 .build());
-        System.out.println("Got reply: " + buyerPartialSignatureMessage);
+        System.out.println("Got reply: " + buyerPartialSignaturesMessage);
 
-        // Buyer sends Message C to seller.
+        // Buyer sends Message C with buyerNonceSharesMessage and buyerPartialSignaturesMessage to seller.
```
stejbac, line 100: The `swapTxInputPartialSignature` field should be redacted from the `buyerPartialSignaturesMessage` field of Message C and held by the buyer in the Bisq Java client until payment, to prevent premature trade closure by the seller. This isn't ideal I guess -- perhaps if it can be ensured that the `GetPartialSignatures` gRPC request is idempotent (which I think it likely currently is) then holding it on the Java end can be avoided.

rpc/src/main/java/bisq/TradeProtocolClient.java
```diff
@@ -127,10 +132,11 @@ private static void testMusigService_twoParties(MusigGrpc.MusigBlockingStub stub
 
         // Buyer sends Message E to seller.
```
stejbac: At this point, the buyer releases the withheld `swapTxInputPartialSignature` field, so it should be included in Message E.

stejbac (commented):
Sorry for the delay responding. I had a look at the comment changes in `TradeProtocolClient` and left some feedback. Coincidentally, I also downgraded the Maven build to JDK 21 in the recently submitted https://github.com/bisq-network/bisq-musig/pull/51. (My changes to `TradeProtocolClient` there will likely also lead to some minor merge conflicts.) Otherwise, the changes look fine.

rpc/src/main/java/bisq/TradeProtocolClient.java
HenrikJannsen: I tried to redact the field at the message but then got some error, but I still dont have a full understanding of the low level protocol yet, so cannot comment more on that...

HenrikJannsen:
> Sorry for the delay responding. I had a look at the comment changes in `TradeProtocolClient` and left some feedback. Coincidentally, I also downgraded the Maven build to JDK 21 in the recently submitted #51. (My changes to `TradeProtocolClient` there will likely also lead to some minor merge conflicts.) Otherwise, the changes look fine.

I guess as you did some of the changes your self (java 21 downgrade) and some of my comments have to be removed again, we can simply drop that PR, then you don't need to deal with merge conflicts.

PR #69, opened 2025-06-26
Upgrade musig2 and BDK dependencies of `rpc` package

stejbac:
Make breaking-change updates of the `musig2`, `bdk_wallet` and `bdk_bitcoind_rpc` crates to 0.3.1, 2.0.0 and 0.20.0 respectively. This involves some minor code changes involving the RNG API and handling of tx eviction from the mempool during RPC syncs.

(The `protocol` and other packages have not been touched as part of this PR, but probably should be changed to use the latest versions as well.)

Also do some minor dependency version updates and tidy the import lists to exactly agree with rustfmt ordering, as they weren't quite in alphabetical order.. (I'm not sure if the entire `rpc` source tree should be formatted with rustfmt instead of Intellij formatting, as this doesn't look like it will give all that great results.)


ChrisSon15:
ACK

Issue #67, opened 2025-06-24
create API for protocol crate


ChrisSon15:
In this article I describe how the interaction with the protocol crate shall work:
https://github.com/bisq-network/bisq-musig/blob/main/concept/ProtocolSubStates.md

First thing to do would be to write an API in the crate protocol to expose the 'signing of Txs' functionality. This functionality can be viewed in the crate protocol, module lib, tests::initial_tx_creation
There are 4 rounds of communication for the complete signing of Txs each of which has a struct Round1Parameter,...
round 1 to 4 together with the parameters should be exposed in the API of crate protocol (may be we can use it as it is).
The the structs shall be exposed via gRPC.
To do that look at crate rpc and use the server, however you probably need to create new gRPC structures.

try to expose the rounds1 to 4 to JAVA.

PR #71, opened 2025-06-27
feat: implement gRPC service for the signing protocol

54corbin:
issue #67


protocol/src/lib.rs
54corbin, line 1: most of the changes on this file are done by  formatter

PR #73, opened 2025-07-02
feat: Implement BMP gRPC signing API

54corbin:
This pull request introduces a gRPC service for the Bisq Musig Protocol (BMP), enabling programmatic interaction with the multi-signature trading protocol.

PR #75, opened 2025-07-21
Musig RPC service security enhancements

stejbac:
This PR changes the Musig gRPC API, together with its mocked-up trade protocol implementation in the `rpc` crate, to test out a few security enhancements:

1. Prevent the buyer from learning the unsigned swap tx during trade setup (unlike all the other prepared txs, where byte-for-byte identical unsigned txs are constructed by each side independently). Instead, the buyer effectively signs the swap tx blindly, using just a sighash supplied by the seller at the trade start. This has the side effect of postponing the buyer's partial-signing of the swap tx by one round, but doesn't otherwise affect the number of rounds.
2. _(Fits in with the above.)_ Redact the buyer's partial signature on the swap tx on the _server_ side, at trade setup, by adding a `buyerReadyToRelease` flag to the `GetPartialSignatures` request proto, instead of expecting the Java client to hold onto sensitive data. This is necessary anyway, because the buyer doesn't know the sighash to sign with until the following round. Now the client is expected to call this method a second time with the flag set.
3. Similarly, add a `sellerReadyToRelease` flag to the `SignSwapTx` request proto, so that the entire response can be redacted when first called, immediately after the buyer confirms payment to check the buyer's partial signature validity. The client should similarly call this method a second time with the flag set, at the trade end.

The main purpose of (1) is to prevent nuisance-broadcast of the swap tx by the buyer, any time after a cooperative trade closure. That is difficult to avoid if the buyer learns the unsigned swap tx, since the adaptor is an affine congruence that necessarily works _both ways_ between the swap tx signature and the buyer payout private key.

As a result of (2) and (3), no cryptographic secrets are being passed to the Java client to secure anymore, so long as they remain sensitive.


ChrisSon15:
re 1.: Isnt it a security risk if the buyer signs the SwapTx blindly? When the seller gives the buyer only the sighash to sign, then the seller could get a signature for any transaction he wants to, which has the same pubkey, in this case that would be Q', see https://github.com/bisq-network/bisq-musig/blob/main/concept/SingleTxOverview.md, the Deposit Tx output 1. Since the seller would get a valid signature for Q' he could trick the buyer into signing the key spend of  Output 0 from `Waning TX Alice`. Since the buyer cant tell the difference because he is signing blindly, he would give the seller the signature for getting those funds. After the trade begins and `DepositTx` is broadcasted,  the seller could immediately broadcast the `Warning Tx Alice`, then immediately broadcast the forged `Swap Tx`which will payout both deposits and the trade amount into his account. That way the seller could steel the funds. Am I missing something here?

Also I am trying to understand what the buyer may gain by knowing the `SwapTx`. Since the input is know anyway, he probably can only learn the output which is just an address created by the Seller. But learning an address is not anything security related or is there anything else to it?

stejbac:
The `xpub` comes in because if the seller is careless and leaks it, anyone who learns it obtains a watch-only view of the seller's entire keychain, as far as I understand. So in particular, they would be able to just guess the address the Swap Tx is spending to. We want to keep that address hidden from the buyer (and unguessable), so that the buyer can't just guess the entire unsigned Swap Tx (through a process of trial-and-error, trying out every known address on the seller's keychain to come up with a Swap Tx candidate that matches the seller-supplied sighash). So the seller would send the sighash to the buyer (to blindly sign), but no other information about the Swap Tx.

The whole point is to somehow stop the buyer from learning the unsigned Swap Tx, since then he would know the signed Swap Tx and be able to broadcast it himself, at the end of the trade (since we can't stop him from learning the signature itself, so some other part of the Swap Tx -- any part, really -- needs to be kept hidden).

ChrisSon15:
> The whole point is to somehow stop the buyer from learning the unsigned Swap Tx, since then he would know the signed Swap Tx and be able to broadcast it himself, at the end of the trade (since we can't stop him from learning the signature itself, so some other part of the Swap Tx -- any part, really -- needs to be kept hidden).

Now I am getting it. Yes that's a great idea to use the blind signature to hide the unsigned SwapTx!

stejbac:
I've rebased and force-pushed to resolve the conflicts with main, occurring after the merge of #73.

(I also inserted a commit at the beginning to do some assorted minor cleanup, prior to the rebase, like removing the now-defunct Cargo lockfiles in the individual workspace members, now that a common project workspace has been created.)

ChrisSon15 (approved):
great

PR #76, opened 2025-08-10
Add mock-protocol tx builders, missing Taproot tweaks, optional trade fee & Claim Txs

stejbac:
Add dedicated tx builders for the mock trade protocol implementation in the `rpc` package (exposed by the `Musig` gRPC service defined in `rpc.proto`), so that real sighashes can be used, instead of just signing hard-coded dummy messages. The payout & anchor addresses are still mocked and hard-coded, as well as the entire Deposit Tx construction, as there is no actual wallet integration at present (and some more thought needs to be put into tx fee handling when merging the buyer+seller input lists and their optional change outputs).

Also update the `Musig` service proto scheme to exchange missing Claim Tx payout addresses, nonces & partial signatures, to build a pair of prepared Claim Txs. And accept an optional trade-fee receiver (address + amount) from the client, to use as an extra output of the Deposit Tx. Also add checks that the _redirection_ receiver amounts sum to the correct total, logging warnings/errors if they don't (but not otherwise failing the trade), and add code+comments to `TradeProtocolClient.java` to illustrate how to properly use the updated API. Take care to modify the gRPC service in a way that won't break the Bisq2 client, though, to give it a chance to stay in sync.

Finally, make sure that all the multisig pubkeys are tweaked, with suitably distinct tap-tweaks, to address the security concerns raised in the comments of #75. In particular, the payout pubkeys/addresses must be distinct from the Warning escrow pubkeys/addresses, so need to be given different tweaks. This can be neatly achieved by providing a script-spend path for each warning escrow output (giving each a non-null merkle root), but making the Deposit outputs keyspend only. To this end, add a _Claim_ script-spend path using `OP_CHECKSEQUENCEVERIFY` to each Warning escrow output, which serves as a backup to the (keyspend-only) prepared Claim Txs, in case of fee volatility (or just for extra payout flexibility).

--

TODO: Consider moving the builders into the `protocol` crate and use them for the implementation there, to help keep the tx construction and wallet interface logic separate from each other and from the crypto, for better modularity and testability. This could be done in a separate PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -646,9 +646,9 @@ dependencies = [
 
 [[package]]
 name = "dyn-clone"
-version = "1.0.19"
+version = "1.0.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1c7a8fb8a9fbf66c1f703fe16184d10ca0ee9d23be5b4436400408ba54a95005"
+checksum = "d0881ea181b1df73ff77ffaaf9c7544ecc11e82fba9b5f27b262a3c73a332555"
 
 [[package]]
 name = "either"
@@ -975,7 +975,7 @@ dependencies = [
  "hyper",
  "libc",
  "pin-project-lite",
- "socket2 0.6.0",
+ "socket2",
  "tokio",
  "tower-service",
  "tracing",
@@ -1400,9 +1400,9 @@ dependencies = [
 
 [[package]]
 name = "prost"
-version = "0.13.5"
+version = "0.14.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2796faa41db3ec313a31f7624d9286acf277b52de526150b7e69f3debf891ee5"
+checksum = "7231bd9b3d3d33c86b58adbac74b5ec0ad9f496b19d22801d773636feaa95f3d"
 dependencies = [
  "bytes",
  "prost-derive",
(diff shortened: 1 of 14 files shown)


ChrisSon15:
The use of a `OP_CHECKSEQUENCEVERIFY` script was considered in the first version of the protocol, then later that was changed. See `concept/update.md` for a discussion of this issue. Did you take the arguments in there into consideration?

stejbac:
I was aware of the privacy argument for using a keyspend-only prepared Claim Tx, instead of a unilaterally & dynamically created Claim Tx making a script spend. (Though I don't really understand the point about being easier to implement, since the technical overhead to add script spend paths, then use them later in a tx built on-the-fly, is surely fairly trivial.)

As I mentioned (in the commit messages and above), the script spend path can be considered a backup and not necessarily the preferred way to make a claim payout. But I think it does no harm to put it in, plus it's fairly important to make sure the Warning Tx outputs are tweaked _differently_ to the Deposit Tx outputs, so there's no address reuse, and a script spend path on one but not the other is a convenient way to do that.

In general, I don't really understand @suddenwhipvapor's concern about blockchain privacy in the sad paths, as they won't be used very often at all. Also, it's surely very hard, if not impossible, to prevent the Warning, Claim & Redirect txs from sticking out on the blockchain at all -- the Warning Tx spends two outputs of the _same tx_, which surely can't be all that common, and the Redirect Tx sends to all the publicly known BM addresses. Plus all the prepared txs are using relative timelocks and I wonder how common they are on the blockchain (compared to absolute timelocks, which are often added by wallets to prevent fee sniping by miners). There may be a fair number appearing in forced Lightning channel closures and other contract-related stuff -- I'm not really sure of the stats (but it's probably worth looking into).

PR #79, opened 2025-08-14
feat: add empty wallet service

54corbin:
Added a skeleton of the wallet gRPC service:
- The bmp_wallet.proto file defining the service(refer to: https://github.com/bisq-network/bisq2/pull/3728 ).
- An implementation of the service in BmpWalletServiceImpl that returns dummy data for testing.
- Integration of the new service into the musigd.


stejbac (approved):
utACK

(I wasn't sure exactly how to test the new wallet service itself against a client, but the code seems fine and it builds OK.)

PR #81, opened 2025-08-22
feat(wallet): BMP wallet crate

sdmg15:


wallet/README.md
ChrisSon15, line 1: please add some info on how to run the tests, even if its standard. To make as easy as possible to actually setup the wallet and run the tests for an outsider.

rpc/src/main/proto/bmp_wallet.proto
```diff
@@ -0,0 +1,149 @@
...
+}
+
+message ListUtxosRequest {}
+
+message ListUtxosResponse {
+    repeated Utxo utxos = 1;
+}
+
+message SendToAddressRequest {
+    optional string passphrase = 1;
```
ChrisSon15: Whats the passphrase for?

rpc/src/main/proto/bmp_wallet.proto
```diff
@@ -0,0 +1,149 @@
...
+    string witness = 5;
+}
+
+message TransactionOutput {
+    uint64 value = 1;
+    string address = 2;
+    bytes scriptPubKey = 3;
+}
+
+message Transaction {
```
ChrisSon15: As stated before, when transferring the transaction to the Java gui. we probably want to send different fields, as the GUI is interested in a different display of information. See Bisq1->Funds->Transaction to learn whats displayed there.

rpc/src/main/proto/rpc.proto
```diff
@@ -99,12 +99,11 @@ message DepositPsbt {
 
 message PublishDepositTxRequest {
   string tradeId = 1;
-  DepositPsbt depositPsbt = 2;
+  DepositPsbt peersDepositPsbt = 2;
```
ChrisSon15: Why do we need this change?

rpc/src/server.rs
```diff
@@ -58,17 +63,23 @@ impl musig_server::Musig for MusigImpl {
...
             trade_model.aggregate_key_shares()?;
-            trade_model.set_trade_amount(Amount::from_sat(request.trade_amount));
-            trade_model.set_buyers_security_deposit(Amount::from_sat(request.buyers_security_deposit));
-            trade_model.set_sellers_security_deposit(Amount::from_sat(request.sellers_security_deposit));
-            trade_model.set_deposit_tx_fee_rate(FeeRate::from_sat_per_kwu(request.deposit_tx_fee_rate));
-            trade_model.set_prepared_tx_fee_rate(FeeRate::from_sat_per_kwu(request.prepared_tx_fee_rate));
+            trade_model.set_trade_amount(
+                Amount::from_sat(request.trade_amount.check_in_signed_range()?));
+            trade_model.set_buyers_security_deposit(
+                Amount::from_sat(request.buyers_security_deposit.check_in_signed_range()?));
```
ChrisSon15: Please dont reformat someone else's code.

wallet/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+        let bmp_wallet = BMPWallet::from_mnemonic("h b c d", Network::Regtest);
+        Ok(())
+    }
+
+    #[test]
+    fn test_load_wallet() -> anyhow::Result<()> {
+        let mut bmp_wallet = BMPWallet::new(Network::Regtest)?;
+        assert!(bmp_wallet.balance() == Amount::from_sat(0));
+
+        let pk1 = PrivateKey::generate(Network::Regtest);
```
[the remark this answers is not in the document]
sdmg15: I can see that in the other crates they're using `Scalar` which defined to be 

```
pub struct Scalar {
    #[cfg(feature = "secp256k1")]
    pub(crate) inner: secp256k1::SecretKey
}
``` 
Which is pretty similar to:

```
pub struct PrivateKey {
    /// Whether this private key should be serialized as compressed
    pub compressed: bool,
    /// The network kind on which this key should be used
    pub network: NetworkKind,
    /// The actual ECDSA key
    pub inner: secp256k1::SecretKey,
}
```
I can change the type name for keeping consistency across the project.

stejbac:
I should be able to carry out a review in the next few hours.

wallet/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+};
+
+use rand::RngCore;
+use std::vec;
+
+const DB_PATH: &str = "bmp_bdk_wallet.db3";
+const ELECTRUM_URL: &str = "ssl://electrum.blockstream.info:6000";
+
+pub struct BMPWallet {
+    wallet: PersistedWallet<rusqlite::Connection>,
```
stejbac, line 18: The use of qualified names like `rusqlite:Connection`, when `Connection` is already in the import list, is giving weak warnings in my IDE (IntelliJ IDEA). Can you change these to either consistently use qualified names throughout the module and drop the import, or consistently use unqualified names? There are also some warnings for `anyhow::Result::Ok`, which is similarly needlessly qualified in a few places, when you could just use `Ok`.

Annoyingly, Clippy doesn't seem to warn about this.

wallet/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+    #[test]
+    fn test_create_wallet_from_mnemonic() -> anyhow::Result<()> {
+        let bmp_wallet = BMPWallet::from_mnemonic("h b c d", Network::Regtest);
+        Ok(())
+    }
+
+    #[test]
+    fn test_load_wallet() -> anyhow::Result<()> {
+        let mut bmp_wallet = BMPWallet::new(Network::Regtest)?;
+        assert!(bmp_wallet.balance() == Amount::from_sat(0));
```
stejbac, line 307: This could use `assert_eq!`.

wallet/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+    };
+
+    fn tear_down() {
+        fs::remove_file(DB_PATH).expect("DB removed after tests");
+    }
+
+    #[test]
+    fn test_create_wallet() -> anyhow::Result<()> {
+        let mut bmp_wallet2 = BMPWallet::new(Network::Regtest)?;
+        assert!(bmp_wallet2.balance() == Amount::from_sat(0));
```
stejbac, line 275: This could use `assert_eq`.

stejbac (changes requested):
I made some styling and other fairly minor suggestions.

More generally, I think it would be nice to made the code a bit more modular by abstracting the persistence and chain sync, as we will probably need to use backends besides Electrum. It seems a bit test-hostile at the moment, especially with hardcoded DB file paths and URLs. But perhaps that could be done in further PRs.

wallet/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+        trx.execute_batch(&Self::schema())?;
+        trx.commit()?;
+
+        Ok(())
+    }
+
+    // Create a new wallet
+    // 1-
+    pub fn new(network: Network) -> anyhow::Result<Self> {
+        // TODO: Make the word size configurable?
```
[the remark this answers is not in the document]
sdmg15: Yes during the creation of the wallet the user of the interface specifies the network

wallet/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+            return Ok(Self {
+                wallet,
+                client,
+                imported_keys,
+                db,
+                network,
+            });
+        }
+
+        return Err(anyhow::anyhow!("Unable to load wallet"));
```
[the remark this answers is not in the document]
ChrisSon15: talking about clippy @sdmg15 did you run clippy an work on the output? I think we should do in the future, and I mean myself foremost ;-)

wallet/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+    pub fn balance(&self) -> Amount {
+        self.wallet.balance().trusted_spendable()
+    }
+
+    // Generate a new bitcoin address, we don't take into account addresses from imported keys
+    pub fn next_unused_address(&mut self) -> AddressInfo {
+        let addr = self.wallet.next_unused_address(KeychainKind::External);
+
+        // Persist the revealed address, to avoid address reuse
+        self.wallet.persist(&mut self.db).expect("Write is okay");
```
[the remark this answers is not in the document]
sdmg15: Will go with a `Result` type for `next_ununused_address` so that the caller maybe able to retry in case of error

PR #84, opened 2025-09-10
SQLCipher + Argon2 Proof of Concept

54corbin:
This POC demonstrates how to use Argon2 to derive a key from a password and use it to encrypt a SQLite database with SQLCipher.


sdmg15:
image

Tested it locally and seems to be working.


Also imported the database into this tool https://sqlite.org/fiddle/index.html. And it errors with `Error: file is not a database`. 

From the POC I can see that once the DB is unlocked it stays so for remaining time. I guess in the final implementation we'll have something like `unlock_for(time)` something similar which is done by bitcoin-core.

poc/sqlcipher_argon2/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+            fs::remove_file(&salt_path)?;
+        }
+
+        // Derive encryption key from password using Argon2
+        let salt = SaltString::generate(&mut OsRng);
+        fs::write(&salt_path, salt.as_str())?;
+
+        let argon2 = Argon2::default();
+        let password_hash = argon2
+            .hash_password(password.as_bytes(), &salt)
```
stejbac, line 111: I think it may be better to use `hash_password_into` directly instead of `hash_password`, as it mentions in the rustdoc that the latter should be used for password-based authentication only, rather than key derivation (for which it says the former should be used). It probably doesn't make much difference in this particular case, as it appears that the password hash is already the required 32 bytes and doesn't actually get truncated, and the two methods appear to me to do the same thing in that case.

I think the default features of `argon2` should accordingly be disabled, probably just using the `alloc` feature, so that the reexported `password-hash` crate isn't used at all, since we're only using `argon2` as a bare-bones KDF, rather than using the [PHC string formatting](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md) functionality provided by the `password-hash` crate.

poc/sqlcipher_argon2/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+            .map_err(|e| anyhow::anyhow!(e.to_string()))?;
+
+        // Extract the hash portion for SQLCipher key (64 hex chars)
+        let output = password_hash.hash.unwrap();
+        let key_bytes = output.as_bytes();
+        let key_hex = hex::encode(&key_bytes[..32]); // Use first 32 bytes (256 bits)
```
stejbac, line 117: This is probably a very minor point, but `hex::encode` doesn't appear to be strictly constant-time, as it's using 16-byte table lookups to encode each 4-bit nibble of the key into a hex character. Truly constant-time byte-array to hex-string conversion could probably be done with some bit bashing -- perhaps that's being a bit too paranoid.

stejbac: I doubt the hex library makes any such security guarantees , as it would make the implementation more complicated for a niche benefit. The implementation doesn't appear to me to be strictly constant-time. OTOH, a 16-byte lookup table likely fits in a single cache line, so it's probably not much of an issue in practice. Plus there's no point hardening it unless the SQLite code which decodes the hex is adopting a similar level of paranoia.

poc/sqlcipher_argon2/src/main.rs
```diff
@@ -0,0 +1,235 @@
...
+        let salt_path = format!("{db_path}.salt");
+        let salt_str = fs::read_to_string(&salt_path)?;
+
+        // Derive the correct key using stored salt
+        let salt =
+            SaltString::from_b64(salt_str.as_str()).map_err(|e| anyhow::anyhow!(e.to_string()))?;
+        let argon2 = Argon2::default();
+        let mut key_bytes = [0u8; 32]; // 256-bit key
+        argon2
+            .hash_password_into(password.as_bytes(), salt.as_str().as_bytes(), &mut key_bytes)
```
stejbac, line 153: I don't think this is right. I think you're meant to use `salt.decode_b4(&mut buf)` instead of using the base64 string itself as the salt.

PR #87, opened 2025-09-23
Move tx builders to `protocol` crate

stejbac:
The following code changes are made by this PR:
1. Apply review suggestions for #83 (add comments, fix wrong sighash type, use named fields in place of tuple struct), that was merged into main pending those changes;
2. Remove the backwards compatibility logic that was added to avoid breaking the Bisq2 client following changes to the `Musig` RPC API, namely addition of Claim Tx-related proto fields, requiring trade-fee/redirection receiver addresses to be on the correct network (Regtest) instead of Signet, and validating the redirection total. These have all been made hard errors instead of just logging a warning/error, now that https://github.com/bisq-network/bisq2/pull/3856 has been merged into main;
3. Move the (Deposit, Swap, Claim, Redirect & Warning) tx builders from the `rpc` crate to the `protocol` crate, where they can hopefully be used to help separate the crypto from the wallet- and tx-related logic.

I also moved the Clippy config of the `rpc` crate into the workspace manifest, so that it can be shared with the `protocol` crate (and possibly the other crates), prior to moving the tx builders there, and disabled warnings module-by-module rather than across the whole `protocol` crate, to continue to get a clean build without losing Clippy coverage.

- ~~**TODO:**~~ Provide a real implementation of the (moved) `protocol::psbt::TradeWallet` trait, instead of just a mock impl (say by making `bdk_wallet::Wallet` implement `TradeWallet`), so that the tx builders can construct a Deposit Tx which spends funds from a real wallet;
- **TODO:** Make the existing end-to-end protocol implementation in `protocol::protocol_musig_adaptor` use the tx builders. (Possibly this should be done in a separate PR.)


stejbac:
I think it may be better to leave both TODOs for a separate PR, as I need to think further about how best to integrate the `TradeWallet` trait with the code in the new `wallet` crate.

stejbac:
I pushed one more commit to address the first TODO, implementing the `TradeWallet` trait for the raw BDK wallet (in addition to the existing mock implementation used by `rpc/protocol.rs`). Further impls could be provided, such as for the `BMPWallet` struct in the new `wallet` crate, but I think I'll leave that for another PR, as well as the second TODO.

ChrisSon15:
utACK

PR #92, opened 2025-10-14
feat(signer): signing inputs own by imported keys and main wallet

sdmg15:


wallet/src/bmp_wallet.rs
```diff
@@ -222,6 +230,51 @@ impl WalletApi for BMPWallet<Connection> {
...
+                let xonly_pubkey = XOnlyPublicKey::from_slice(&xonly_pubkey)
+                    .expect("Should be valid xonlypub key");
+                let script = ScriptBuf::new_p2tr(secp, xonly_pubkey, None);
+
+                if script == *input_script {
+                    return (true, Some(*key));
+                }
+            }
+            (false, None)
+        };
```
ChrisSon15: Why do i need that bool in the return if I have the Option as second paramter?

wallet/src/bmp_wallet.rs
```diff
@@ -222,6 +230,51 @@ impl WalletApi for BMPWallet<Connection> {
...
+    fn sign(
+        &mut self,
+        psbt: &mut Psbt,
+        sign_options: SignOptions,
+    ) -> anyhow::Result<(), SignerError> {
+        let secp = self.secp_ctx();
+        let is_mine = |input_script: &ScriptBuf| {
+            for key in &self.imported_keys {
+                let xonly_pubkey = key.base_point_mul().serialize_xonly();
+                let xonly_pubkey = XOnlyPublicKey::from_slice(&xonly_pubkey)
```
ChrisSon15: Probably not super important now. but if we will have more key in the DB. Everytime we need to pubkey to do the multiplication is probably some performance overhead. Something to keep an eye on.

sdmg15: Yes indeed. It's done this way to avoid IO. Cause the other way of doing this will be loading each wallet from disk and then toggle the signing.

PR #94, opened 2025-10-21
Integrate tx builders into end-to-end protocol

stejbac:
Replace most of the tx and PSBT building and signing logic in `protocol::protocol_musig_adaptor` with the functionality provided by the tx builders in `protocol::transaction`, recently moved from the `rpc` crate. Add an appropriate builder field to each of the `protocol_musig_adaptor::(Deposit|Swap|Warning|Redirect|Claim)Tx` structures, removing the `tx` fields (and `*psbt` fields in the case of the Deposit Tx) and holding the constructed signed/unsigned txs/PSBTs in the builder fields instead. (The `WarningTx::key_spend: Option<AggKey>` field was also removed, as it was stored but not used -- I'm not sure whether some future purpose for that field was intended, but it would be easy to put it back in.)

The crypto in `protocol_musig_adaptor` remains largely untouched by this PR, though it does some minor factoring out of the crypto from the tx-related code.

To help ensure that all the txs of the end-to-end protocol have the exact specified feerate (currently hardcoded to 20 sat/vB for the Deposit Tx and 10 sat/vB for all the other protocol txs), the algorithm to filter and calculate the absolute receiver amounts from their shares was ported from the Bisq2 Java codebase, in the method `protocol::receiver::Receiver::conpute_receivers_from_shares`, giving the option to calculate the receiver list on either side. (This uncovered a couple of minor bugs in the Java version of the algorithm.)

Aside from the feerates, the only other changes to the resultant protocol txs are making all their inputs opt-in RBF and shuffling all the inputs and outputs of the Deposit Tx.

--

The PR also canonicalises the import grouping of all the `rpc` & `protocol` crate files according to the recently added `StdExternalCrate` rustfmt config setting, for consistency across the project.

(To resolve import list conflicts in `protocol_musig_adaptor.rs`, I also did a merge from main.)

### DIFF
--- a/protocol/src/protocol_musig_adaptor.rs
+++ b/protocol/src/protocol_musig_adaptor.rs
@@ -7,10 +7,9 @@ use bdk_wallet::bitcoin::hashes::sha256t::Hash;
 use bdk_wallet::bitcoin::key::Secp256k1;
 use bdk_wallet::bitcoin::taproot::Signature;
 use bdk_wallet::bitcoin::{
-    Address, Amount, FeeRate, KnownHrp, Network, OutPoint, Psbt, PublicKey, ScriptBuf,
-    TapSighashTag, Transaction, TxOut, Txid, relative,
+    Address, Amount, FeeRate, Network, OutPoint, Psbt, ScriptBuf, TapSighashTag, Transaction,
+    TxOut, Txid, XOnlyPublicKey, relative,
 };
-use bdk_wallet::miniscript::ToPublicKey as _;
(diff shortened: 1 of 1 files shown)


stejbac:
I've just rebased the last few commits to modify some of unit test values slightly, for better coverage, and insert a new commit (just before the merge from main) to fix a bug in the receiver amounts calculation fn that can cause total receiver overpayment in some edge cases. (That fix needs to be ported over to the Java side as well, I think.)

ChrisSon15 (approved):

PR #101, opened 2025-11-15
feat: Implement encryption and decryption for wallet database file

sdmg15:
This PR adds encryption and decryption to the wallet interface. Allowing users to optionally encrypt their wallet.
When encrypting a previous plain database wallet we use the `sqlcipher_export` function to migrate the wallet content from plain to encrypted database.

### DIFF
--- a/wallet/src/bmp_wallet.rs
+++ b/wallet/src/bmp_wallet.rs
@@ -1,7 +1,8 @@
 use std::ops::{Deref, DerefMut};
-use std::vec;
+use std::{fs, vec};
 
-use base64::{Engine, engine::general_purpose};
+use base64::engine::general_purpose;
+use base64::Engine;
 use bdk_wallet::bitcoin::bip32::Xpriv;
 use bdk_wallet::bitcoin::hex::DisplayHex;
 use bdk_wallet::bitcoin::{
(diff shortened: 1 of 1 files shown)


wallet/src/bmp_wallet.rs
```diff
@@ -378,8 +382,17 @@ impl WalletApi for BMPWallet<Connection> {
 
     // For already created wallets this will load stored data
     // This will also load the imported keys
-    fn load_wallet(network: Network) -> anyhow::Result<Self> {
-        let mut db = Connection::open(Self::DB_PATH)?;
+    fn load_wallet(network: Network, password: Option<&str>) -> anyhow::Result<Self> {
+        let mut db = if let Some(password) = password {
+            let salt = get_salt(Self::DB_PATH)?;
```
ChrisSon15: Does this mean, that the DB cannot change its location on the file system?

sdmg15: It could be better named db_name cause it actually doesn't encodes the path. The current implementation uses the current execution directory.

ChrisSon15: That would mean that the salt is constant for all users. And that would defeat the purpose of a salt.
The salt should be ideally be generated from secure random. And needs to be stored a long side the database. However its not a secret. It ensures that a rainbow attack is inefficient, by making then same password used for 2 different databases result in 2 different encryption keys. Here is what you need to do:

1. generate the salt from random (like 128 bits or more)
2. store it in a file next to the database (files doesn't need to be encrypted)
3. when decrypting read the salt.
4. password change or any re-encrypting leads to re randomizing the salt. (Do we have that case at all?)

PR #102, opened 2025-11-15
Move protocol integration tests

stejbac:
Move the end-to-end protocol tests out of `protocol/src/lib.rs` and into their own file under `protocol/tests`, which is the standard place for integration tests. Don't move the `nigiri` module out of the production crate as it's currently also being used by `rpc::bmp_service`.

Also add a semaphore to make them run in series (`tokio::sync::Semaphore`, which was chosen mainly because it is fair, which hopefully reduces the tendency to get warnings that a given test has been running for >60s). A permit is supplied every time `nigiri::check_start` is called and released at the end of the given test.

Slightly modify the test instructions in the README, accordingly.

Additionally, update the project dependencies. Exclude `bdk_wallet` from the update, however, and pin it to 2.1.* for now, as the latest 2.2.0 release is giving deprecation warnings from the use of `bdk_wallet::wallet::signer` without providing a clear workaround. (I suppose, alternatively, the warnings could just be suppressed, as their own example code is doing, but I don't think there's anything new in the current 2.2 release that would be helpful to us.)


54corbin:
> Additionally, update the project dependencies. Exclude **`bmp_wallet`** from the update, however, and pin it to 2.1.* for now

you mean `bdk_wallet`  right?

PR #103, opened 2025-11-17
some code improvement for testenv

ChrisSon15:


testenv/src/lib.rs
```diff
@@ -73,23 +75,17 @@ impl TestEnv {
 
     /// Mine blocks using bitcoind RPC
     pub fn mine_blocks(&self, count: usize) -> Result<Vec<BlockHash>> {
-        // Generate blocks to the node's mining address
-        let mining_address_str = self.bitcoind.client.get_new_address(None, None)?.0;
-        let mining_address: Address<NetworkUnchecked> = mining_address_str.parse()?;
-        let mining_address_checked = mining_address.require_network(Network::Regtest)?;
         let block_hashes = self
             .bitcoind
             .client
```
54corbin: Would you think we should further extract `bitcoind.client` as a field inside the `TestEnv` as it's beening using everywhere.

54corbin:
ack

PR #118, opened 2026-01-29
fix: wallet signing keys after reload/encrypt/decrypt

### LINKED ISSUE #117: Wallet: after reload it not possible to sign
I have noticed while working on #116 that when you load the wallet and try to create and sign transactions with it the wallet doesn't have the signing keys loaded.

sdmg15:
This PR fixes an issue that happened when the wallet was reloaded or decrypted. The signing keys weren't loaded with the wallet.
The fix applied here is that we only now load the signing keys into the wallet when `sign` is called. Closes #117


ChrisSon15:
ACK

PR #121, opened 2026-02-18
Feat transition rpc to testenv

ChrisSon15:


wallet/src/bmp_wallet.rs
```diff
@@ -1078,8 +1071,7 @@ mod tests {
     #[test]
     #[should_panic = "value: file is not a database"]
     fn encrypted_wallet() {
-        let _permit = SEMAPHORE.acquire();
-        let _tmp_dir = tear_up();
+        let env = TestEnv::new().unwrap();
```
sdmg15: I think the Testenv should be used only in the integration tests? Because this will uselessly download bitcoind and electrs which aren't needed for the unit tests.
If we absolutely need to use the Testnenv we should extract the only needed part I think.

PR #122, opened 2026-02-19
feat: Initialize tracing crate

sdmg15:
This PR add a common tracing module that will be used across the project;

### DIFF
--- a/rpc/src/wallet.rs
+++ b/rpc/src/wallet.rs
@@ -5,18 +5,18 @@
 
 use std::sync::{Arc, Mutex, RwLock};
 
-use bdk_bitcoind_rpc::bitcoincore_rpc::{Client, RpcApi as _};
 use bdk_bitcoind_rpc::Emitter;
+use bdk_bitcoind_rpc::bitcoincore_rpc::{Client, RpcApi as _};
 use bdk_wallet::bitcoin::{Network, Transaction, Txid};
 use bdk_wallet::chain::{ChainPosition, CheckPoint, ConfirmationBlockTime};
 use bdk_wallet::{AddressInfo, Balance, KeychainKind, LocalOutput, Wallet};
-use bmp_tracing::tracing::{debug, error, info, trace};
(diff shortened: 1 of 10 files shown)


rpc/src/wallet.rs
```diff
@@ -7,13 +10,13 @@ use bdk_bitcoind_rpc::Emitter;
 use bdk_wallet::bitcoin::{Network, Transaction, Txid};
 use bdk_wallet::chain::{ChainPosition, CheckPoint, ConfirmationBlockTime};
 use bdk_wallet::{AddressInfo, Balance, KeychainKind, LocalOutput, Wallet};
+use bmp_tracing::tracing::{debug, error, info, trace};
```
stejbac: I think that tracing/logging in production library crates (as opposed to binary and test crates) should only be using a facade and importing `tracing` macros directly, rather than re-exported via the new `bmp_tracing` crate, and the library crate itself (ideally) shouldn't have a dependency on `bmp_tracing`, only on the `tracing`/`log` facade crate.

(Library crates should ideally be written so that they could be reused in different projects, that could be using a completely independent tracing/logging subscriber.)

.github/workflows/cont_integration.yml
```diff
@@ -65,7 +65,7 @@ jobs:
         run: cargo clippy --all-targets
       
       - name: Unit and Integration tests
-        run: cargo test
+        run: RUST_LOG=off cargo test
```
stejbac: Maybe it would be better if the default is not to log (uncaptured) when running `cargo test`, so that the console output is clean, unless a user requests log output. (I'm not sure exactly how other Rust projects that have test logging subscribers handle this.)

rpc/src/wallet.rs
sdmg15: Thanks for checking this. I've also thought the same.
Please take a look at https://github.com/bisq-network/bisq-musig/pull/122/commits/6935bdadfff1ffa2d4a744fc80bf83c8962fbcec if I'm missing something

PR #124, opened 2026-02-25
feat: add draining wallet API

sdmg15:
This adds a new function to the wallet API `drain` which will generate a new receiving address from the the internal wallet and send all the funds from imported keys to it.

The fees are computed from the total amount of imported balance to be sent.


ChrisSon15 (commented):
Some minor stuff

PR #146, opened 2026-04-08
made TestEnv able to run the tests in parallel, which reduced the tes…

ChrisSon15:
…t time on my end from 4m to 1m.

Also remove unneeded tmp_dir

### DIFF
--- a/test-runner.sh
+++ b/test-runner.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+# this is for continuous test under multithreaded conditions.
+export RUST_BACKTRACE=full
+export TEST_MULTITHREADED=true
+
+for i in {32..1}; do
+    echo "Running tests with $i threads..."
+    cargo test -- --test-threads=$i
+    if [ $? -ne 0 ]; then
+        echo "Tests failed with $i threads. Exiting."
+        exit 1
+    fi
+done
+
+echo "All tests passed for threads 1 to 32."
--- a/testenv/src/lib.rs
+++ b/testenv/src/lib.rs
@@ -1,6 +1,7 @@
 //! Bitcoin regtest environment using electrsd with automatic executable downloads
 
 use std::net::SocketAddrV4;
+use std::sync::{Mutex, MutexGuard, PoisonError};
 use std::time::Duration;
 
 use anyhow::{Context as _, Result};
@@ -24,7 +25,7 @@ use tempfile::TempDir;
 use tokio::net::TcpListener;
 
 /// Bitcoin regtest environment manager
-pub struct TestEnv {
+pub struct TestEnv<'a> {
     bitcoind: Node,
     electrsd: ElectrsD,
     timeout: Duration,
@@ -36,6 +37,7 @@ pub struct TestEnv {
     explorer_port: Option<u16>,
     bitcoin_rpc_pwd: String,
     mempool: Vec<Txid>,
+    _guard: Option<MutexGuard<'a, ()>>,
 }
 
 /// Configuration parameters.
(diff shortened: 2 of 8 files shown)


testenv/src/lib.rs
```diff
@@ -33,8 +31,7 @@ pub struct TestEnv {
     delay: Duration,
     bdk_electrum_client: BdkElectrumClient<Client>,
     ctx: Secp256k1<All>,
-    _permit: Permit,
-    _tmp_dir: TempDir,
+    //  _permit: Permit,
```
stejbac: This line should be deleted if no longer needed, rather than commented out.

testenv/src/lib.rs
```diff
@@ -196,7 +192,8 @@ impl TestEnv {
 
     /// create environment with automatic downloads
     pub fn new_with_conf(config: Config) -> Result<Self> {
-        let permit = SEMAPHORE.acquire(); // have testenvs single threaded because of bitcoind and electrs references.
+        // let permit = SEMAPHORE.acquire(); // have testenvs single threaded because of bitcoind
```
stejbac: This line should be deleted if no longer needed, rather than commented out.

testenv/src/lib.rs
```diff
@@ -256,8 +253,7 @@ impl TestEnv {
             delay: config.delay,
             bdk_electrum_client,
             ctx: Secp256k1::new(),
-            _permit: permit,
-            _tmp_dir: tmp_dir,
+            // _permit: permit,
```
stejbac: This line should be deleted if no longer needed, rather than commented out.

testenv/src/lib.rs
```diff
@@ -196,7 +192,8 @@ impl TestEnv {
 
     /// create environment with automatic downloads
     pub fn new_with_conf(config: Config) -> Result<Self> {
-        let permit = SEMAPHORE.acquire(); // have testenvs single threaded because of bitcoind and electrs references.
+        // let permit = SEMAPHORE.acquire(); // have testenvs single threaded because of bitcoind
+        // and electrs references.
         let tmp_dir = tempdir().expect("failed to create temporary directory");
         std::env::set_current_dir(tmp_dir.path()).expect("failed to set current directory");
```
stejbac: I don't see how setting the current directory can possibly work in parallel, without being protected by a global mutex/semaphore held by whatever is relying on it being a particular path.

I really don't think anything (other than possibly some very top level application startup code) should care about the current directory, and it should be completely unaffected by its value. If that's not the case, we should change it.

--

Furthermore, after removing the `_tmp_dir` field from the struct, the `tmp_dir` local variable will be immediately dropped after having set the current directory to it, so the current directory will be pointing to an invalid/stale path.

stejbac (changes requested):
In addition to the in-code suggestions, we should remove `simple-semaphore` from the Cargo manifest, as it is no longer a crate dependency after these changes.

sdmg15 (changes requested):
Image

Image

Tests are flaky. This is related to the comment made by stejbac.

testenv/src/lib.rs
ChrisSon15: The tmp_dir was a left over not used.
Temp directories now need to be created when needed.
thanks for spotting

stejbac (approved):
utACK

(`simple-semaphore` still needs to be removed from the `testenv` Cargo manifest, though.)

sdmg15 (approved):
tAck

test-runner.sh
sdmg15, line 1: Not sure you really want to push this?

ChrisSon15: yes, but will delete it soon when no longer needed. So its just temporarly

PR #148, opened 2026-04-10
fix: hand over new addresses until stop gap reached

sdmg15:
Closes #139

### DIFF
--- a/wallet/src/bmp_wallet.rs
+++ b/wallet/src/bmp_wallet.rs
@@ -177,6 +177,7 @@ impl BMPWalletPersister for Connection {
     }
 }
 
+const STOP_GAP: usize = 50;
 #[allow(unused)]
 pub struct BMPWallet<P: BMPWalletPersister> {
     wallet: PersistedWallet<P>,
@@ -188,13 +189,19 @@ pub struct BMPWallet<P: BMPWalletPersister> {
 
 impl BMPWallet<Connection> {
     pub fn next_address(&mut self, key_chain: KeychainKind) -> anyhow::Result<AddressInfo> {
-        // FIXME: `next_unused_address` just returns the same unused address over and over. It has
-        //  to either be marked as used (which change isn't staged and therefore presumably never
-        //  persisted) or a fresh address requested with `reveal_next_address`.
-        let addr = self.wallet.next_unused_address(key_chain);
+        let unused = self.list_unused_addresses(key_chain).collect::<Vec<_>>();
+
+        let addr = if unused.len() >= STOP_GAP {
+            unused
+                .first()
+                .expect("Unused addresses should not be empty")
+                .clone()
+        } else {
+            self.next_unused_address(key_chain)
+        };
+
         // Persist the revealed address to avoid address reuse
         self.persist()?;
-
         Ok(addr)
     }
 
--- a/wallet/src/bmp_wallet.rs
+++ b/wallet/src/bmp_wallet.rs
@@ -192,12 +192,14 @@ impl BMPWallet<Connection> {
         let unused = self.list_unused_addresses(key_chain).collect::<Vec<_>>();
 
         let addr = if unused.len() >= STOP_GAP {
-            unused
-                .first()
+            let rand_index = rand::random::<u8>() as usize % unused.len();
+
+            println!("Rand {rand_index}");
+            unused.get(rand_index)
                 .expect("Unused addresses should not be empty")
                 .clone()
         } else {
-            self.next_unused_address(key_chain)
+            self.reveal_next_address(key_chain)
         };
 
         // Persist the revealed address to avoid address reuse
@@ -767,6 +769,7 @@ impl DerefMut for BMPWallet<Connection> {
 #[cfg(test)]
 mod tests {
     use std::str::FromStr;
+
     use bdk_kyoto::FeeRate;
     use bdk_wallet::bitcoin::hashes::Hash as _;
     use bdk_wallet::bitcoin::{Address, AddressType, Amount, BlockHash, Network, Weight, psbt};
@@ -776,9 +779,9 @@ mod tests {
     use bmp_tracing::tracing;
     use rand::RngCore as _;
     use secp::Scalar;
-    use tempfile::{tempdir, TempDir};
+    use tempfile::{TempDir, tempdir};
 
-    use crate::bmp_wallet::{BMPWallet, WalletApi as _};
+    use crate::bmp_wallet::{BMPWallet, STOP_GAP, WalletApi};
     use crate::test_utils::{MockedBDKElectrum, derive_public_key, load_imported_wallet};
 
     fn get_dir() -> TempDir {
@@ -1167,6 +1170,31 @@ mod tests {
 
         assert_eq!(w1.balance(), Amount::from_int_btc(1));
         assert_eq!(w2.balance(), Amount::ZERO);
+        Ok(())
+    }
+
+    #[test]
+    fn test_address_generation() -> anyhow::Result<()> {
+        let dir = get_dir();
+        let mut wallet = BMPWallet::new(dir.path(), "", Network::Regtest)?;
+
+        let mut add_vec: Vec<AddressInfo> = vec![];
+
+        loop {
+            add_vec.push(wallet.next_address(KeychainKind::External)?);
+            if add_vec.len() >= STOP_GAP {
+                break;
+            }
+        }
+
+        // Since we reached STOP_GAP, next_address should return one address from the add_vec list
+        let new_addr = wallet.next_address(KeychainKind::External)?;
+        assert!(add_vec.contains(&new_addr));
+
+        // Returned next address should be different from previous one but still exist in the list
+        let new_addr2 = wallet.next_address(KeychainKind::External)?;
+        assert!(add_vec.contains(&new_addr2) && new_addr != new_addr2);
+
         Ok(())
     }
 }
--- a/wallet/src/bmp_wallet.rs
+++ b/wallet/src/bmp_wallet.rs
@@ -1,3 +1,4 @@
+use std::collections::VecDeque;
 use std::ops::{Deref, DerefMut};
 use std::path::Path;
 use std::{fs, vec};
@@ -23,6 +24,7 @@ use bdk_wallet::{
     TxOrdering, Utxo, Wallet, WalletPersister, WeightedUtxo,
 };
 use rand::RngCore as _;
+use rand::prelude::IndexedRandom as _;
 use secp::Scalar;
 
 use crate::chain_data_source::ChainDataSource;
@@ -178,6 +180,7 @@ impl BMPWalletPersister for Connection {
 }
 
 const STOP_GAP: usize = 50;
+
 #[allow(unused)]
 pub struct BMPWallet<P: BMPWalletPersister> {
     wallet: PersistedWallet<P>,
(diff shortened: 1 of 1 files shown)


stejbac:
The trouble is that this is a fairly simplistic fix as it stands. Once there are at least `STOP_GAP` unused external addresses, `next_address` will keep returning the same unused external address over and over. I imagine it will be quite common for a trader to do a long string of trades without funding the wallet meanwhile, in which case only change addresses (from the HD keychains) will actually be used (assuming all the trades follow the happy path). There will likely be a permanent deficit of used external addresses, as the number of revealed addresses runs ahead of the number of used addresses, always with a difference of `STOP_GAP` (or just under).

It might not be possible to do much better without persisting extra data recording the reserved address indices, but the relevant limit is the biggest _contiguous_ length of unused addresses rather than the total number, so that might give a way to improve the algorithm.

wallet/src/bmp_wallet.rs
```diff
@@ -188,13 +189,19 @@ pub struct BMPWallet<P: BMPWalletPersister> {
...
     pub fn next_address(&mut self, key_chain: KeychainKind) -> anyhow::Result<AddressInfo> {
-        // FIXME: `next_unused_address` just returns the same unused address over and over. It has
-        //  to either be marked as used (which change isn't staged and therefore presumably never
-        //  persisted) or a fresh address requested with `reveal_next_address`.
-        let addr = self.wallet.next_unused_address(key_chain);
+        let unused = self.list_unused_addresses(key_chain).collect::<Vec<_>>();
+
+        let addr = if unused.len() >= STOP_GAP {
+            unused
+                .first()
```
ChrisSon15: this would reuse the first address in that list, which ever that is. to reduce contention on one address reuse please randomly select one address?

ChrisSon15 (changes requested):

ChrisSon15:
> The trouble is that this is a fairly simplistic fix as it stands. Once there are at least `STOP_GAP` unused external addresses, `next_address` will keep returning the same unused external address over and over. I imagine it will be quite common for a trader to do a long string of trades without funding the wallet meanwhile, in which case only change addresses (from the HD keychains) will actually be used (assuming all the trades follow the happy path). There will likely be a permanent deficit of used external addresses, as the number of revealed addresses runs ahead of the number of used addresses, always with a difference of `STOP_GAP` (or just under).
> 
> It might not be possible to do much better without persisting extra data recording the reserved address indices, but the relevant limit is the biggest _contiguous_ length of unused addresses rather than the total number, so that might give a way to improve the algorithm.

In the case of a trader doing many trades one after the other: For each trade the protocol will request several addresses. Once the trade ends is a happy path, the output of the DepositTx will be spend on the next Trades DepositTx (since we have a coin selection strategy of spending trade outputs first). Then all the unused Addresses which are in the old trade presigned transactions cannot be send to the chain anymore. 
In case of an unhappy end of the trade,we would see these addresses on-chain and the wallet will give out new addresses in future.
So in case of chaining the trades, I don't see that address-reuse will appear on chain. 

Only  if the trader makes multiple trades in parallel and at least 2 parallel trades end up in an unhappy path, then we could see address reuse on chain. Lets say we have 5 requested addresses per trade (didn't count it), then we need round about 10 concurrent trades of which at least 2 go into unhappy path. In that case the STOP_GAP of 50 could be exceeded and address reuse could happen on chain. I guess this is unlikely.

The fact that the wallet is handing out the same address several times does not mean they appear on chain. And if unused addresses are given to different trade partners (but never used) I don't see a problem in that.

wallet/src/bmp_wallet.rs
```diff
@@ -192,12 +192,14 @@ impl BMPWallet<Connection> {
         let unused = self.list_unused_addresses(key_chain).collect::<Vec<_>>();
 
         let addr = if unused.len() >= STOP_GAP {
-            unused
-                .first()
+            let rand_index = rand::random::<u8>() as usize % unused.len();
+
+            println!("Rand {rand_index}");
+            unused.get(rand_index)
                 .expect("Unused addresses should not be empty")
```
ChrisSon15: Since we have a Result as retrun type anyway, I guess we should remove 'expect' so reviewer do tip over it in the future.

wallet/src/bmp_wallet.rs
```diff
@@ -1169,4 +1172,29 @@ mod tests {
...
+            if add_vec.len() >= STOP_GAP {
+                break;
+            }
+        }
+
+        // Since we reached STOP_GAP, next_address should return one address from the add_vec list
+        let new_addr = wallet.next_address(KeychainKind::External)?;
+        assert!(add_vec.contains(&new_addr));
+
+        // Returned next address should be different from previous one but still exist in the list
```
ChrisSon15, line 1219: When reusing addresses, you select addresses by random, so its not gurateed, that the new_addr2 != new_addr Its just a chance.

ChrisSon15 (changes requested):

wallet/src/bmp_wallet.rs
```diff
@@ -23,6 +24,7 @@ use bdk_wallet::{
     TxOrdering, Utxo, Wallet, WalletPersister, WeightedUtxo,
 };
 use rand::RngCore as _;
+use rand::prelude::IndexedRandom;
```
stejbac: Should that be `use rand::prelude::IndexedRandom as _;`?

wallet/src/bmp_wallet.rs
```diff
@@ -177,23 +179,69 @@ impl BMPWalletPersister for Connection {
...
 
 impl BMPWallet<Connection> {
     pub fn next_address(&mut self, key_chain: KeychainKind) -> anyhow::Result<AddressInfo> {
-        // FIXME: `next_unused_address` just returns the same unused address over and over. It has
-        //  to either be marked as used (which change isn't staged and therefore presumably never
-        //  persisted) or a fresh address requested with `reveal_next_address`.
-        let addr = self.wallet.next_unused_address(key_chain);
-        // Persist the revealed address to avoid address reuse
-        self.persist()?;
+        let mut unused = self.list_unused_addresses(key_chain).collect::<Vec<_>>();
```
ChrisSon15: Claude code says: 'Wallet::list_unused_addresses(keychain) returns all revealed addresses that have never received a transaction output — not just those after the last on-chain-seen address.`
This is unfortunately not the behaviour we expected. We would need here the list of all unused addresses since the last address on chain.

wallet/src/bmp_wallet.rs
```diff
@@ -177,23 +179,69 @@ impl BMPWalletPersister for Connection {
...
+
+        let addr = if unused.len() >= STOP_GAP {
+            // Use proper random selection and collision detection
+            let mut selected_addr: Option<AddressInfo> = None;
+            let mut attempts = 0;
+            const MAX_ATTEMPTS: usize = 5;
+            const RECENT_ADDRESS_CACHE_SIZE: usize = 10;
```
ChrisSon15: we discussed something different in last teammeeting, and right now I dont see what made you use this algorithm. the benefits are not obvious to me.
What i expected at this point: Have a index into the ordered-set of unused-addresses and increase it index by 1 when a (reused) address is given out.
This 'index' could be the address given out itself, when looking for the next address, we could search the old-address in the unused addresses and take the next in the ordered-set.

ChrisSon15 (changes requested):
Please see my changes.
Also i would like to discuss about the algorithm used. As its not intuitiv maybe at least a comment explaining it would be good.

ChrisSon15:
Subject: [PATCH] wallet: only hand out addresses past the last on-chain-used index
---
Index: wallet/src/bmp_wallet.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/wallet/src/bmp_wallet.rs b/wallet/src/bmp_wallet.rs
--- a/wallet/src/bmp_wallet.rs	(revision a3e3290dd25774d77a6f8cedc4a41db4b849c30a)
+++ b/wallet/src/bmp_wallet.rs	(revision 716f9ca5a33c1c48ac50aee4eb29cb12b252ac5b)
@@ -192,8 +192,29 @@
 }
 
 impl BMPWallet<Connection> {
+    /// Returns all revealed-but-unused addresses with a derivation index
+    /// strictly greater than the highest index that has been observed on chain
+    /// (i.e. for which a `TxOut` has been indexed).
+    ///
+    /// In contrast to [`Wallet::list_unused_addresses`], which returns *every*
+    /// revealed address that has never received an output — including "gap"
+    /// addresses with a lower index than ones already used — this only returns
+    /// addresses revealed after the last on-chain activity. If no address on
+    /// `key_chain` has been seen on chain yet, all revealed unused addresses
+    /// are returned.
+    pub fn list_unused_addresses_since_last_used(
+        &self,
+        key_chain: KeychainKind,
+    ) -> impl Iterator<Item = AddressInfo> + '_ {
+        let last_used = self.spk_index().last_used_index(key_chain);
+        self.list_unused_addresses(key_chain)
+            .filter(move |info| last_used.is_none_or(|idx| info.index > idx))
+    }
+
     pub fn next_address(&mut self, key_chain: KeychainKind) -> anyhow::Result<AddressInfo> {
-        let mut unused = self.list_unused_addresses(key_chain).collect::<Vec<_>>();
+        let mut unused = self
+            .list_unused_addresses_since_last_used(key_chain)
+            .collect::<Vec<_>>();
 
         let addr = if unused.len() >= STOP_GAP {
             // Use proper random selection and collision detection
@@ -1213,6 +1234,109 @@
         assert_eq!(w2.balance(), Amount::ZERO);
         Ok(())
     }
+
+    #[test]
+    fn test_list_unused_addresses_since_last_used() -> anyhow::Result<()> {
+        let dir = get_dir();
+        let mut wallet = BMPWallet::new(dir.path(), "", Network::Regtest)?;
+
+        // Reveal a handful of addresses (indices 0..=4).
+        let revealed: Vec<AddressInfo> = (0..5)
+            .map(|_| wallet.reveal_next_address(KeychainKind::External))
+            .collect();
+        let revealed_indices: Vec<u32> = revealed.iter().map(|a| a.index).collect();
+        assert_eq!(revealed_indices, vec![0, 1, 2, 3, 4]);
+
+        // With no on-chain activity, the new method returns the same set
+        // as list_unused_addresses.
+        let baseline: Vec<u32> = wallet
+            .list_unused_addresses(KeychainKind::External)
+            .map(|a| a.index)
+            .collect();
+        let since_last: Vec<u32> = wallet
+            .list_unused_addresses_since_last_used(KeychainKind::External)
+            .map(|a| a.index)
+            .collect();
+        assert_eq!(since_last, baseline);
+        assert_eq!(since_last, vec![0, 1, 2, 3, 4]);
+
+        // Receive an output on the address at index 2. The "last used" index
+        // is now 2, so list_unused_addresses_since_last_used must exclude
+        // indices 0 and 1 (gap addresses) even though list_unused_addresses
+        // still includes them.
+        receive_output_to_address(
+            &mut wallet,
+            revealed[2].address.clone(),
+            Amount::ONE_BTC,
+            ReceiveTo::Block(chain::ConfirmationBlockTime {
+                block_id: BlockId {
+                    height: 2,
+                    hash: BlockHash::all_zeros(),
+                },
+                confirmation_time: 2,
+            }),
+        );
+
+        let baseline: Vec<u32> = wallet
+            .list_unused_addresses(KeychainKind::External)
+            .map(|a| a.index)
+            .collect();
+        // The unfiltered list still surfaces the gap indices 0 and 1.
+        assert_eq!(baseline, vec![0, 1, 3, 4]);
+
+        let since_last: Vec<u32> = wallet
+            .list_unused_addresses_since_last_used(KeychainKind::External)
+            .map(|a| a.index)
+            .collect();
+        // Filtered list drops everything at or below the last used index.
+        assert_eq!(since_last, vec![3, 4]);
+
+        // Receive on the highest revealed index (4). No unused address with
+        // a greater index exists yet, so the iterator should be empty —
+        // confirming the threshold tracks the *maximum* used index.
+        receive_output_to_address(
+            &mut wallet,
+            revealed[4].address.clone(),
+            Amount::ONE_BTC,
+            ReceiveTo::Block(chain::ConfirmationBlockTime {
+                block_id: BlockId {
+                    height: 3,
+                    hash: BlockHash::all_zeros(),
+                },
+                confirmation_time: 3,
+            }),
+        );
+
+        let since_last: Vec<u32> = wallet
+            .list_unused_addresses_since_last_used(KeychainKind::External)
+            .map(|a| a.index)
+            .collect();
+        assert!(
+            since_last.is_empty(),
+            "expected no unused addresses past the highest used index, got {since_last:?}"
+        );
+
+        // Reveal one more address; it should now be the sole result.
+        let new_addr = wallet.reveal_next_address(KeychainKind::External);
+        assert_eq!(new_addr.index, 5);
+
+        let since_last: Vec<u32> = wallet
+            .list_unused_addresses_since_last_used(KeychainKind::External)
+            .map(|a| a.index)
+            .collect();
+        assert_eq!(since_last, vec![5]);
+
+        // The internal keychain has had no on-chain activity at all, so the
+        // method must fall back to "all revealed unused" for that keychain.
+        let internal_addr = wallet.reveal_next_address(KeychainKind::Internal);
+        let internal_since_last: Vec<u32> = wallet
+            .list_unused_addresses_since_last_used(KeychainKind::Internal)
+            .map(|a| a.index)
+            .collect();
+        assert!(internal_since_last.contains(&internal_addr.index));
+
+        Ok(())
+    }
 
     #[test]
     fn test_address_generation() -> anyhow::Result<()> {

wallet/src/bmp_wallet.rs
sdmg15: Can you enlighten me about what will be the difference?

sdmg15:
@ChrisSon15 @stejbac simplified the algo based on suggestions https://github.com/bisq-network/bisq-musig/pull/148/changes/2dd8499575fb39989326b9890fb84b73c6e6ba25, please take a look

PR #150, opened 2026-05-08
Crate Chain & Wallet integration

ChrisSon15:
created a new crate chain
created Testchain for out current test related development 
integrated it into protocol crate
by introducing a trait for broadcasting transactions
integrated it into wallet crate
by introducing a trait for syncing wallets


chain/src/lib.rs
```diff
@@ -0,0 +1,35 @@
+use std::sync::Arc;
+
+use bdk_wallet::bitcoin::{Transaction, Txid};
+use bdk_wallet::chain::spk_client::{FullScanRequest, FullScanResponse};
+
+/// Abstraction over blockchain interaction for broadcasting transactions.
+/// to be extended
+pub trait ChainApi: Send + Sync {
+    fn transaction_broadcast(&self, tx: &Transaction) -> anyhow::Result<Txid>;
```
sdmg15, line 9: Nit: Could be naming here be simplified to something like `broadcast_tx`?

chain/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+/// performing a full keychain scan. Wallet-level sync routines compose these primitives instead
+/// of taking a hard dependency on a specific chain client (e.g. `BdkElectrumClient`).
+pub trait ChainScanner {
+    /// Insert transactions into the backend's transaction cache so it will not re-fetch them.
+    /// Typically used to pre-populate the cache from an existing `TxGraph` before a scan.
+    fn populate_tx_cache(&self, txs: impl IntoIterator<Item = impl Into<Arc<Transaction>>>);
+
+    /// Full scan the keychain scripts described by `request` against the chain backend and return
+    /// updates suitable for applying to a `bdk_wallet` data structure.
+    fn full_scan<K: Ord + Clone>(
```
sdmg15, line 22: Since we're going to use CBF, the method signature doesn't quite match.
For CBF with kyoto we'll need params like `peers`, `scanType` (Sync vs Recovery).
Should we add another method here?

protocol/tests/protocol_integration_tests.rs
```diff
@@ -42,16 +39,20 @@ fn initial_tx_creation(env: &mut TestEnv) -> anyhow::Result<(BMPProtocol, BMPPro
     let alice_funds = funded_wallet(env);
     let bob_funds = funded_wallet(env);
 
+    let alice_client = Box::new(env.new_testchain()?);
+    let bob_client = Box::new(env.new_testchain()?);
```
sdmg15, line 67: If this creates two separate client  don't we need to make sure they're connected together? So that when a transaction is broadcasted they both receive when syncing?

chain/src/lib.rs
ChrisSon15: Once we (or you) move to CBF, then we introduce a change. For now I am glad if we can get this straighten out..

protocol/tests/protocol_integration_tests.rs
ChrisSon15: The client should not be connected together, the server (Electrum or CBF) should sync the clients, so the transaction is broadcasted from one client to the other.

chain/src/lib.rs
ChrisSon15: I'd rather stay with it as is.

protocol/tests/protocol_integration_tests.rs
```diff
@@ -1,16 +1,36 @@
...
+    dead_code,
+    reason = "one variant is always unused depending on the WALLET_BACKEND constant"
+)]
+enum WalletBackend {
+    Mem,
+    Bmp,
+}
+
+/// Single switch controlling which wallet backend the integration tests exercise.
+const WALLET_BACKEND: WalletBackend = WalletBackend::Mem;
```
sdmg15: Can't we make this as argument or read from env if nothing is provided just use mem instead of having an enum as it's easy for this line to be modified and committed many times by inadvertence.

protocol/tests/protocol_integration_tests.rs
```diff
@@ -33,17 +61,36 @@ pub fn funded_wallet(env: &mut TestEnv) -> MemWallet {
     wallet
 }
 
+fn funded_bmp_wallet(env: &mut TestEnv) -> BMPWallet<Connection> {
+    // The wallet persists to disk; the integration tests don't need the directory to outlive
+    // them, but we use `keep()` so the test does not race the wallet against `TempDir`'s
+    // automatic cleanup.
+    let dir = tempfile::tempdir().unwrap().keep();
```
sdmg15: I'm not seeing where we're removing this directory since we have `keep()`.

wallet/src/bmp_wallet.rs
```diff
@@ -330,12 +332,33 @@ impl ProtocolWalletApi for BMPWallet<Connection> {
         is_selected: &dyn Fn(&OutPoint) -> bool,
     ) -> anyhow::Result<()> {
         let mut psbt_copy = psbt.clone();
-        self.sign(&mut psbt_copy, SignOptions::default())?;
+        // Trade-deposit PSBTs have their `tap_key_origins` and friends redacted before they're
+        // exchanged, so we need to repopulate the bip32 derivation paths before BDK can finalize
+        // the inputs we own. Also tell BDK to trust the witness UTXO since the full previous
+        // transactions are not carried in the PSBT.
+        self.wallet
```
sdmg15: Can we have tests for this `sign_selected_inputs`?

wallet/src/protocol_wallet_api.rs
stejbac, line 49: Can we please make this `&self`, rather than changing the rest of the code to use `&mut self` for `sign_selected_inputs` (unless there's a specific reason why some of the implementations need to mutate the wallet just to sign a PSBT)? BDK `Wallet::sign` takes `&self`, and I think we should do the same in our own signing methods.

protocol/tests/protocol_integration_tests.rs
ChrisSon15: Good point, will use env

protocol/tests/protocol_integration_tests.rs
ChrisSon15: in latest version I have redone the TestEnv's creating temp dirs entirely to have the lifetimes of the tempdirs correctly and have the tempdirs cleanup.

wallet/src/bmp_wallet.rs
ChrisSon15: It doesn't have any test coverage yet. I wasn't aware that you used a new Wallet for customPayout. So wallets in use, we have now:
- BMPWallet
- MemWallet
- Wallet (BDK)
- test_utils::get_funded_wallet_single(desc)  [underlying BDK_Wallet]
- MockTradeWallet

I guess we have to wait for the test case 'bdk_trade_wallet_multisig_signing' to be ported to BMPWallet (or MemWallet).

wallet/src/protocol_wallet_api.rs
stejbac: OK, I see that `BMPWallet::sign` takes `&mut self`, so that it can lazily load the wallet signers, as the xprv isn't persisted by BDK Wallet and thus the wallet starts out watch-only when it's first loaded from disc. I suppose the signers could be preloaded, but delaying the loading of the wallet secrets until they're actually needed seems like a reasonable design choice.

So maybe it's better to keep it as a mutating method for the time being. (I guess interior mutability isn't really an option as it would force all accesses to the underlying BDK wallet to be synchronised.)

sdmg15 (approved):
We have both `MemWallet::funded_wallet` and `funded_mem_wallet` which from my understading both do the same thing. 
A part from that I like the new abstraction especially the BoxedTradeWallet and ChainApi. the wallet just now need to adapt better with the changes.

protocol/src/psbt.rs
```diff
@@ -309,40 +337,53 @@ impl TradeWallet for Wallet {
         psbt.redact_sensitive_fields();
         Ok(psbt)
     }
+}
 
-    fn sign_selected_inputs(&self, psbt: &mut Psbt, is_selected: &dyn Fn(&OutPoint) -> bool) -> Result<()> {
-        if !is_well_formed(psbt) {
-            return Err(TransactionErrorKind::InvalidPsbt);
+impl TradeWallet for BMPWallet<Connection> {
+    fn new_internal_key(&mut self) -> Result<XOnlyPublicKey> {
```
stejbac: This method is duplicating the logic in the `bdk_wallet::Wallet` impl, which we should avoid.

Also, it uses `reveal_next_address` to get a new address+index from which to determine the corresponding internal key, whereas it should be using `BMPWallet::next_address`.

I think `new_internal_key` should be a method of the `ProtocolWalletAPI` base trait rather than `TradeWallet`, since it should always return the internal key whose corresponding P2TR address equals the result of calling `ProtocolWalletAPI::new_address` instead, so the two methods are closely linked.

protocol/src/psbt.rs
stejbac: I've just noticed the `MemWallet` impl of `create_half_deposit_psbt` is broken. It should be pushing the half-deposit placeholder SPK + deposit-amount into the vector, _then_ the trade fee receivers (with the correct respective amounts, _not_ `deposit_amount`). It just happened to work because `trade_fee_receivers` is always empty wherever `MemWallet` is currently being used.

The logic in `create_half_deposit_psbt` is already being duplicated in three places with this PR, which is very error prone. I think it should be a default method of `TradeWallet`, overridden only in `MockTradeWallet`. The original BDK `Wallet` version is no longer needed, as it can call out to `ProtocolWalletAPI::create_psbt` just like the `MemWallet` and `BDKWallet` versions, to get identical behaviour (excluding error types).

protocol/Cargo.toml
```diff
@@ -22,6 +21,7 @@ wallet = { path = "../wallet" }
 bdk_wallet = { workspace = true, features = ["test-utils"] }
 bmp_tracing = { workspace = true }
 const_format = { workspace = true }
+tempfile = { workspace = true }
```
stejbac: Where is this dependency being used in `protocol`?

rpc/src/protocol.rs
```diff
@@ -764,17 +766,22 @@ pub enum ProtocolErrorKind {
     MissingTradeWallet,
     #[error("missing script key")]
     MissingScriptKey,
-    #[error("insufficient redirection funds (available {available_msat:?} msat, used {used_msat:?} msat)")]
+    #[error(
+        "insufficient redirection funds (available {available_msat:?} msat, used {used_msat:?} msat)"
+    )]
```
stejbac, line 771: Would it possible to avoid unnecessary reformatting of attributes like this? My IDE autoformatter doesn't like the way Rustfmt is trying to format it, creating a tug-of-war. (Plus I think a single line looks better.)

rpc/src/protocol.rs
```diff
@@ -764,17 +766,22 @@ pub enum ProtocolErrorKind {
...
     InsufficientRedirectionFunds {
         available_msat: u64,
         used_msat: u64,
     },
-    #[error("excess redirection funds (available {available_msat:?} msat, used {used_msat:?} msat)")]
+    #[error(
+        "excess redirection funds (available {available_msat:?} msat, used {used_msat:?} msat)"
+    )]
```
stejbac, line 778: Same as above.

Cargo.toml
```diff
@@ -37,7 +38,8 @@ tail_expr_drop_order = { level = "allow", priority = 1 }
 edition_2024_expr_fragment_specifier = { level = "allow", priority = 1 }
 
 [workspace.lints.clippy]
-pedantic = "warn"
+# `priority = -1` lets per-lint entries below override the group default.
+pedantic = { level = "warn", priority = -1 }
```
stejbac: We should set the priorities consistently, rather than a mixture of 1 vs 0 and 0 vs -1.

Cargo.toml
```diff
@@ -46,7 +48,7 @@ iter_on_empty_collections = "warn"
 iter_on_single_items = "warn"
 missing_const_for_fn = "warn"
 renamed_function_params = "warn"
-significant_drop_tightening = "warn"
+significant_drop_tightening = { level = "allow" }
```
stejbac: I investigated and found that it is quite easy to prevent this lint from firing, by just dropping the `TestEnv` instance at the bottom of each method body where it is being declared, so I think we should keep this lint as "warn".

stejbac (changes requested):

Cargo.toml
ChrisSon15: Well, i knew this already before checking in the code. When devs write new code,they usually dont think about when to drop Testenv. then clippy comes in and suggests to drop it too early. If the dev goes with that warning he introduces a bug. So the clippy hint is actually pushing the new dev to produce a difficult to find bug (which happend to me). Therefore I think this clippy warning is a bad idea.

Cargo.toml
ChrisSon15: Claude code refuses, I think its right:

Summary

  Result: the requested change is not possible — and the user's if possible qualifier covered exactly this case. Cargo refuses it.

  What I tried

  Changed [workspace.lints.clippy] to use the same "1 vs 0" pattern as [workspace.lints.rust]:

  pedantic = "warn"               # implicit priority 0 (matches the group lints in rust section)
  missing_errors_doc = { level = "allow", priority = 1 }   # overrides at 1

  Why it failed

  cargo clippy errors out:

  Cargo.toml:42:1: error: lint group `pedantic` has the same priority (0) as a lint

  Cargo requires that a lint group's priority differs from any individual lint's priority in the same scope. The clippy section has many sibling lints at implicit priority 0 (allow_attributes,
  branches_sharing_code, exhaustive_enums, iter_on_*, missing_const_for_fn, renamed_function_params, str_to_string, try_err, unused_trait_names, use_self), so pedantic can't sit at 0 alongside them.

  The [workspace.lints.rust] section escapes this only because rust-2024-compatibility is the lone group and its sibling unused-qualifications isn't a member of that group.

  Why it works in the rust section

  The two sections genuinely differ: the rust section has no group/non-group priority collision, so 1 vs 0 is fine; the clippy section does, so the group must sit at -1 to give the non-group lints room
  at 0.

  Action

  Reverted Cargo.toml to the original pedantic = { level = "warn", priority = -1 }. Verified cargo clippy --workspace succeeds. No files left modified by this task.

rpc/src/protocol.rs
ChrisSon15: I am not sure with of the tools I used touched those lines. Probably claude code. Once touched it will be reformatted with Rustfmt. Once formatted, I have a hard time changing it back to its original, because if I change the line it will be reformatted again. So I would need to change my setup only to bring back any desired 'non-Rustfmt' style.
I guess we need one uniform formatting style.

protocol/Cargo.toml
ChrisSon15: oops, thats a left over.

protocol/src/psbt.rs
ChrisSon15: yes, very good point.

rpc/src/protocol.rs
stejbac: If the IDE can't be configured not to mangle the formatting of nearby code, just use a bare-bones text editor to fix up any such unwanted changes and then commit the fix-up manually from the command line (say using `git commit -a --amend --no-edit` if it was the previous commit that included unwanted changes).

My IDE (IntelliJ IDEA Community Edition) allows me to select/deselect the individual hunks of the patch when making a commit, so it's very easy not to include any spurious line changes that weren't intended to be a part of it, but it shouldn't be all that much harder to do the same thing from the command line (say using `git add -p`).

Cargo.toml
stejbac: I don't understand what the AI is talking about and I really don't think we should be paying too much attention to it, let alone allowing it to dictate how the code is written.

It may be right that a lint group can't sit at the same priority as the individual lints within it, but `significant_drop_tightening` is in the `nursery` group (so `allow` by default), not the `pedantic` group. So there should be no need to change the priority of the `pedantic` group to -1, just to delete the line:
```
significant_drop_tightening = "warn"
```

In this way, the Clippy lints section would be as follows:
```
[workspace.lints.clippy]
pedantic = "warn"
# Enable selected 'nursery' and 'restriction' lints...
allow_attributes = "warn"
branches_sharing_code = "warn"
exhaustive_enums = "warn"
iter_on_empty_collections = "warn"
iter_on_single_items = "warn"
missing_const_for_fn = "warn"
renamed_function_params = "warn"
str_to_string = "warn"
try_err = "warn"
unused_trait_names = "warn"
use_self = "warn"
# Probably counterproductive to document errors of all publicly visible fns before the code has matured:
missing_errors_doc = { level = "allow", priority = 1 }
# Panics shall generally only result from bugs in callee code, so this gives too many false positives:
missing_panics_doc = { level = "allow", priority = 1 }
# Probably shouldn't annotate every (public) pure fn with '#[must_use]', so too many false positives:
must_use_candidate = { level = "allow", priority = 1 }
```

Cargo.toml
stejbac: OK, then perhaps it's better to keep `significant_drop_tightening` at its default `allow` level, rather than making it `warn`. I think the lint exists mainly to help with performance anyway, rather than writing correct code.

rpc/src/protocol.rs
ChrisSon15: We have the the script install-hooks.sh in the repo, I am using it. So git will run it, no matter which tool I use. It will reformat the code with every commit. If I want to change that, I need to deinstall the git hook and reinstall it later. cumbersome.
We are having these discussions a couple of times. The way the industry solves this problem is usually to use a formatter everyone agrees on and have that run over then code. Then there are no more ambiguities. 
I cant help it that occasionally there is unnecessary reformatting. If we want to be pedantic about code format then lets introduce a common code formatter, everyone uses. I dont care which one. Otherwise even a change in the comment can trigger a reformat.

Cargo.toml
ChrisSon15: Oh, AI was assumig the lints following the 'pedantic' are all in the pedantic group. Nevermind, changed it as you suggested

rpc/src/protocol.rs
stejbac: I see. Then in that case, I don't think we should be doing any reformatting as part of the pre-commit hook at all, since it was never agreed that we would use rustfmt throughout the project, or even that it would eventually be migrated to use rustfmt throughout.

Every commit ought to give rise to a clean build, under all circumstances that I can think of, so it's reasonable to include Clippy in the pre-commit hook. (We should probably run `cargo check` as well, because not all the `warn`-level lints are Clippy lints.) But if we're not trying to migrate everything over to use rustfmt, I don't think that even modified lines should have it forcibly applied to them.

If there's nothing wrong with the formatting of the original source file (or anything else about it) but the tool won't allow one to fully undo one's changes to it, then the tooling is getting in the way. It should be sanity checking the files, not mangling them involuntarily.

ChrisSon15: Back then we decided to have a pre-comit hook for formatting, if people wanted it. Sdmg15 has programmed it and it was used be sdmg15,corbin and now me. If we decide the pre-commit hook is now a bad idea, we should do this in fridays teammeeting including sdmg15.

Let's move the disucssion there.

Also we should consider cases, where someone renames variables or corrects comments. In these cases the pre-commit hook will reformat a lot of source code even though there is no substantial change. That would destroy your personel formatting as well and give the code base 2 different styles within the same file.

wallet/src/bmp_wallet.rs
```diff
@@ -336,12 +340,33 @@ impl ProtocolWalletApi for BMPWallet<Connection> {
         is_selected: &dyn Fn(&OutPoint) -> bool,
     ) -> anyhow::Result<()> {
         let mut psbt_copy = psbt.clone();
-        self.sign(&mut psbt_copy, SignOptions::default())?;
+        // Trade-deposit PSBTs have their `tap_key_origins` and friends redacted before they're
+        // exchanged, so we need to repopulate the bip32 derivation paths before BDK can finalize
```
stejbac: I don't think this comment is quite correct, as the metadata isn't really redacted. It's not added in the first place, as already discussed.

stejbac (changes requested):
The changes look quite good, but there's still some problematic duplication. Namely, the logic in the `sign_selected_inputs` method is triplicated for the `Wallet`, `BMPWallet<Connection>` and `MemWallet` impls. Would it be possible to factor it out into a `pub(crate)` utility fn or similar, to deduplicate, as was done for `new_internal_key`?

PR #151, opened 2026-05-10
Custom payouts for mediation part 3: add HD keychain support & integration test

stejbac:
Ensure Custom Payout Tx signing works with HD wallet keychains, and not just wallets with a single private key. To this end, add a `protocol_wallet_api::WalletExt` extension trait with a single `update_psbt_with_derivation_paths` method, to help BDK Wallet locate the private keys of any pubkeys found in the tapscripts of the inputs to sign. Without this, the `Wallet::sign` method will only work when the wallet is created from a definite descriptor with a single private key. Update the `sign_selected_inputs` method impls to augment the PSBT metadata with the missing derivation paths, prior to calling `Wallet::sign`.

Also add an integration test to confirm that the signed Custom Payout Txs are broadcastable. This requires updating the protocol implementation in `protocol_musig_adaptor.rs` to exchange multisig script keys (from the traders' respective wallets) with the peer, and adding script paths to the Deposit payout addresses, which were formerly keyspend only. This brings it closer to sync with the protocol implementation in `rpc::protocol` used by the `Musig` gRPC service, which had already been updated similarly in #134.

--

Additionally, make all the `protocol::script_paths::*` public functions return `Result`s, with consistent input validation, to prevent panics in `script_paths::deposit_payout_descriptor` when the buyer & seller multisig keys are identical.

Finally, add a dust limit check to `CustomPayoutTxBuilder` (as well as the other tx builders, which could hypothetically produce dust outputs in the case of very low trade amounts and high tx fees), and ensure that custom payouts pay the correct tx fee even when the payout addresses aren't P2TR (which might occur in the future if they're used to make emergency payouts).

--

This PR hopefully completes the work needed to support mediated custom payouts on the Rust side, leaving just the issue that the BMP wallet code needs to be updated to accept the _external_ private key of the trader's P2TR payout, upon trade completion, instead of the internal key as it does currently.

### DIFF
--- a/protocol/src/script_paths.rs
+++ b/protocol/src/script_paths.rs
@@ -1,34 +1,42 @@
 use bdk_wallet::bitcoin::opcodes::all::{OP_CHECKSIG, OP_CHECKSIGVERIFY, OP_CSV};
 use bdk_wallet::bitcoin::taproot::TaprootBuilder;
 use bdk_wallet::bitcoin::{ScriptBuf, TapNodeHash, XOnlyPublicKey, relative, script};
-use bdk_wallet::miniscript::{DefiniteDescriptorKey, Descriptor};
+use bdk_wallet::miniscript::{DefiniteDescriptorKey, Descriptor, Miniscript, Tap};
 
-use crate::transaction::NetworkParams;
+use crate::transaction::{NetworkParams, Result};
 
-pub fn deposit_payout_merkle_root(buyer_pub_key: &XOnlyPublicKey, seller_pub_key: &XOnlyPublicKey) -> TapNodeHash {
+pub fn deposit_payout_merkle_root(
+    buyer_pub_key: &XOnlyPublicKey,
+    seller_pub_key: &XOnlyPublicKey,
+) -> Result<TapNodeHash> {
     single_path_merkle_root(multisig_script(buyer_pub_key, seller_pub_key))
 }
 
-pub fn warning_escrow_merkle_root(claim_pub_key: &XOnlyPublicKey, network: impl NetworkParams + Copy) -> TapNodeHash {
+pub fn warning_escrow_merkle_root(
+    claim_pub_key: &XOnlyPublicKey,
+    network: impl NetworkParams + Copy,
+) -> Result<TapNodeHash> {
     single_path_merkle_root(claim_script(claim_pub_key, network.claim_lock_time()))
 }
 
-// FIXME: Can panic; improve:
 pub fn deposit_payout_descriptor(
     internal_key: &XOnlyPublicKey,
     buyer_pub_key: &XOnlyPublicKey,
     seller_pub_key: &XOnlyPublicKey,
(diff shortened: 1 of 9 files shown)


protocol/src/script_paths.rs
```diff
@@ -1,34 +1,42 @@
...
 
-// FIXME: Can panic; improve:
 pub fn deposit_payout_descriptor(
     internal_key: &XOnlyPublicKey,
     buyer_pub_key: &XOnlyPublicKey,
     seller_pub_key: &XOnlyPublicKey,
-) -> Descriptor<DefiniteDescriptorKey> {
-    format!("tr({internal_key},and_v(v:pk({buyer_pub_key}),pk({seller_pub_key})))").parse().unwrap()
+) -> Result<Descriptor<DefiniteDescriptorKey>> {
+    Ok(format!("tr({internal_key},and_v(v:pk({buyer_pub_key}),pk({seller_pub_key})))").parse()?)
```
ChrisSon15, line 27: Why not v:pk({seller_pub_key})?

ChrisSon15, line 27: Why not v:pk({seller_pub_key}) ?

wallet/src/protocol_wallet_api.rs
```diff
@@ -224,6 +225,41 @@ impl MemWallet {
...
 
+pub trait WalletExt {
+    fn update_psbt_with_derivation_paths(&self, psbt: &mut Psbt);
+}
+
+impl WalletExt for Wallet {
+    fn update_psbt_with_derivation_paths(&self, psbt: &mut Psbt) {
+        for input in &mut psbt.inputs {
+            for key in input.tap_key_origins.keys() {
+                let spk = ScriptBuf::new_p2tr(&*LIBSECP256K1_CTX, *key, None);
```
ChrisSon15, line 236: Why is there a None, isn#t it about the custom payout, they have a Some(..) merkle_root right?

stejbac: The wallet holds a hash-map of the ScriptPubKeys of revealed (and lookahead, I believe) external addresses to their derivation indices, in the `bdk_chain::spk_txout::SpkTxOutIndex` struct. So to find the derivation index of an internal key, it must first be converted back to the (keyspend-only) wallet address/SPK that it corresponds to.

I don't think the wallet exposes any way to look up an index from an internal key directly, only from an SPK.

protocol/src/script_paths.rs
stejbac: The multisig tapscript is:
```
<buyer_pub_key> OP_CHECKSIGVERIFY <seller_pub_key> OP_CHECKSIG
```
The miniscript fragment `v:pk(key)` translates to `<key> OP_CHECKSIGVERIFY`, whereas `pk(key)` translates to `<key> OP_CHECKSIG`. The miniscript fragment `and_v(X,Y)` is just the concatenation `[X] [Y]` of the translated script fragments `[X]` and `[Y]`.

Issue #147, opened 2026-04-09
use TestEnv for TestServers in rpc


ChrisSon15:
the sub-project rpc user nigiri still and manual installation of bitcoind for testing.

The goal is to replace it with TestEnv. while keepng TetEnv in the test-section of rust. which means if spinning up a bitcoind server is needed we can do it using 'cargo test <some-test>' and not by calling a binary directly. But this should be ok, as this is really a test-environment. Also add either a commandline parameter or environment parameter to read the datadirectory from, if its not given use a new TempDir. A tempDir get deleted after end of server, the datadirectory not.

So a typical Java test that tests the gRPC should do the following:
Start TestEnv (including explorer) by calling a rust test (probably using "cargo test 'some-test'")
Start the Rust gRPC Server (Cargo run --bin musigd)
load the Java rRPC test client
sync-ing on the server by checking if they started already, so we dont have flacky testenvionment when timing is differnt on other computers.

probably the current nigiri.java file needs to be translated to use RPC calls to the TestEnv.

Then 
- rewrite the existing Java test to use this environment.
- remove nigiri
- change the README.md accordingly

stejbac:
For end-to-end testing with Bisq2, which I imagine will be almost entirely manual, we will ultimately need a persistent Regtest bitcoind/electrs instance, since the Alice and Bob Bisq2 instances will be persisting past trades (which need to be watched for breaches) and need a persistent wallet each. I don't think TestEnv will ultimately be suitable for that, as it's meant for creating an ephemeral Regtest setup.

Right now, there is no persistence implemented for the protocol and its trade state, so a persistent Alice/Bob wallet datadir (and persistent Regtest setup shared by Alice & Bob) is of limited usefulness, but we will ultimately need that, I think. Perhaps an ephemeral TestEnv setup could be used meanwhile,

Currently I'm hampered by the fact that during manual testing with Bisq2, the Alice & Bob Rust servers are running in isolation against their respective Bisq2 clients (using mock trade wallets), and there is no shared chain notification mechanism I can think of, even just as a mock, so that one trader sees another trader's (mocked or real) tx broadcasts. This is going to hamper the development of the sad paths of the protocol.

[...]

PR #155, opened 2026-05-21
feat: replace nigiri with testenv rpc

sdmg15:
Created this to quickly gather feedback.
The new way of running the rpc will be: 

1- Run the bitcoind client:

`cargo run --bin testenv-server -p testenv -- --data-dir "$DATA_DIR"`

2- From the output you'll be able to get the RPC node connection information to be used next

```sh
cargo run --bin musigd --manifest-path rpc/Cargo.toml -- \
  --port 50051 \
  --bitcoin-rpc-url=http://127.0.0.1:37459 \
  --bitcoin-rpc-user=bitcoin \
  --bitcoin-rpc-pass=passs
```
3- `mvn clean verify `

But there is an issue: inside `MemWallet::funded_wallet` on line 33 we're launching another testenv in order to fund the wallet. Why not just reusing the same wallet that's created inside `musigd.rs` main function (created while instantiating `WalletImpl`) that way we would be able to fund it inside `musigd.rs`


rpc/src/bin/musigd.rs
```diff
@@ -9,29 +10,83 @@ use rpc::pb::bmp_protocol::bmp_protocol_service_server::BmpProtocolServiceServer
...
+        {
+            Auth::UserPass(user.clone(), pass.clone())
+        } else {
+            // Try cookie file as fallback
+            Auth::CookieFile(std::path::PathBuf::from("~/.bitcoin/.cookie"))
+        };
+
+        BitcoinCoreClient::new(rpc_url, auth)?
+    } else {
+        panic!("Can't proceed without bitcoin_rpc_url set")
```
stejbac: `main()` is already returning a `Result`, so there's no need to panic here.

I think we should avoid panicking in production code, even `main()` fns.

rpc/src/test/java/bisq/BmpServiceIntegrationTest.java
```diff
@@ -21,19 +21,60 @@ public class BmpServiceIntegrationTest {
...
+    /**
+     * Wait for a gRPC service to be ready by attempting to call a method
+     */
+    private void waitForGrpcReady(BmpProtocolServiceGrpc.BmpProtocolServiceBlockingStub stub, int maxAttempts) {
+        for (int i = 0; i < maxAttempts; i++) {
+            try {
+                // Try to ping by calling getBalance with a dummy request
+                System.out.println("  Checking gRPC server readiness (attempt " + (i + 1) + "/" + maxAttempts + ")");
+                // For now, we'll just wait - a better implementation would actually probe
+                Thread.sleep(500);
```
stejbac: Should use Guava `Uninterruptibles.sleepUninterruptibly` I think. Then the try-catch blocks can be avoided. (Plus it's not really good practice to just swallow `InterruptedException`.)

rpc/src/bin/musigd.rs
sdmg15: Yes as said I'm a little bit puzzled whether the rpc crate is there only for testing end-to-end or it'll go in prod too, as from reading the code I can see a mix of tests stuffs into main code

rpc/src/test/java/bisq/TestEnvClient.java
```diff
@@ -0,0 +1,344 @@
...
+
+        String result = rpcCall(
+                "getblockcount",
+                "[]"
+        );
+
+        try {
+            return Integer.parseInt(extractResult(result));
+        } catch (Exception e) {
+            return -1;
```
stejbac: Why do we have to default here, rather than just letting the `RuntimeException` bubble up?

rpc/src/test/java/bisq/TestEnvClient.java
```diff
@@ -0,0 +1,344 @@
...
+
+        String result = rpcCall(
+                "getbalance",
+                "[]"
+        );
+
+        try {
+            return Double.parseDouble(extractResult(result));
+        } catch (Exception e) {
+            return 0.0;
```
stejbac: Why do we have to default here, rather than just letting the `RuntimeException` bubble up?

rpc/pom.xml
```diff
@@ -19,6 +19,14 @@
             <artifactId>netty-codec-http2</artifactId>
             <version>4.1.119.Final</version>
         </dependency>
+
+        <!-- Bitcoin RPC Client -->
+        <dependency>
+            <groupId>org.bitcoinj</groupId>
```
stejbac: Is it really necessary to pull in such a huge dependency? (I assume it's just for address validation, which shouldn't be very important for tests, or is there more to it?)

ChrisSon15:
I think ideally, there would be a test, which starts the 2 server mentioned in 1.+2. in one command for testing purpose.

sdmg15:
> I think ideally, there would be a test, which starts the 2 server mentioned in 1.+2. in one command for testing purpose.

That'd be doable but for debugging this will make it a little bit difficult as the logs will be scattered.

rpc/src/bin/musigd.rs
stejbac: There was never meant to be a mix of test code and production code, when I first wrote `musigd` (and the `rpc` crate). Testenv really doesn't belong anywhere in the production code (and neither did Nigiri -- the latter crept in with the addition of `BMPService`, which is legacy code as of now).

`musigd` was intended to be used in production, with the gRPC port, bitcoind datadir, port, etc. passed as command-line arguments.

stejbac:
I don't think we should bother with fully automated tests for `BMPService` (at least against a Java client). It's really legacy code, as of now, not being used for anything important. `WalletService` is also legacy code not being used for anything important, so not too much further effort should be poured into it. The `Musig` gRPC service _is_ being used by the Bisq2 client, but it's currently only using a mock trade wallet.

I'm not sure it's really worth the trouble of writing fully automated _Java_ tests of any of the gRPC services, running through `cargo test` at least. I think `cargo test` should probably just be used to run pure Rust tests. We already have enough of a problem with slow tests, and firing up a Java client as part of a test (or set of tests), in addition to a bitcoind/electrsd instance, is IMO too slow and complicated a setup for an ordinary integration test started with `cargo test`. I think the Java stuff should be kept largely segregated and considered example code as much as actual test code, since bisq-musig is after all a Rust project.

ChrisSon15:
> I'm not sure it's really worth the trouble of writing fully automated _Java_ tests of any of the gRPC services, running through `cargo test` at least. I think `cargo test` should probably just be used to run pure Rust tests. We already have enough of a problem with slow tests, and firing up a Java client as part of a test (or set of tests), in addition to a bitcoind/electrsd instance, is IMO too slow and complicated a setup for an ordinary integration test started with `cargo test`. I think the Java stuff should be kept largely segregated and considered example code as much as actual test code, since bisq-musig is after all a Rust project.

I agree, that we dont need 'fully automated _Java_ tests of any of the gRPC services, running through `cargo test` at least.' It would indeed slow down the testing and not add much value.
However, for the java developers we need a way to easily fire up a rust backend to test against. For the java devs to have any meaning full test they will need the rpc backend as well as a test bitcoin server. And since its is very easy to start a TestEnv server  and the rpc server in a seperate process, i think we should provide that for java developers. Our Java code in bisq-musig should only show case how the java devs can fireup a Bitcoin server and and rpc server and connect them. The best is if we have one command to fire up both. If the java tests want to fire up the infrastructure for each test or only once and have many tests against the same servers is up to them. Probably it will be a mixture.
Anyway, i wrote the issue https://github.com/bisq-network/bisq-musig/issues/147 to support java devs when writing code against the rpc interface, so they can have automated tests. I think this is important as there will be quite some java code which is neither GUI related or trivial.
So I think we should provide a fully automated Java test of the rRPC services, but it should run standalone as java test only.

ChrisSon15:
> I don't think we should bother with fully automated tests for `BMPService` (at least against a Java client). It's really legacy code, as of now, not being used for anything important. `WalletService` is also legacy code not being used for anything important, so not too much further effort should be poured into it. The `Musig` gRPC service _is_ being used by the Bisq2 client, but it's currently only using a mock trade wallet.
> 

I deleted  `WalletService`  in my latest still not merged PR. If 'BMPService' causing work, we should consider dropping it.

ChrisSon15:
> > I think ideally, there would be a test, which starts the 2 server mentioned in 1.+2. in one command for testing purpose.
> 
> That'd be doable but for debugging this will make it a little bit difficult as the logs will be scattered.

Not sure if I understand your point. If the user has to start the 2 servers themselves, then the logs will be scattered as well, right? I was just suggesting to have one command start both servers as convenience mthod. That doesn't change anything else, or am I missing something?

stejbac:
> I deleted `WalletService` in my latest still not merged PR. If 'BMPService' causing work, we should consider dropping it.

I think we might be talking about a different `WalletService`. I was referring to `rpc::wallet::WalletService` and its gRPC endpoints, which I wrote a long time ago while I was still working on the wallet code, rather than `protocol::wallet_service::WalletService`.

sdmg15:
> > > I think ideally, there would be a test, which starts the 2 server mentioned in 1.+2. in one command for testing purpose.
> > 
> > 
> > That'd be doable but for debugging this will make it a little bit difficult as the logs will be scattered.
> 
> Not sure if I understand your point. If the user has to start the 2 servers themselves, then the logs will be scattered as well, right? I was just suggesting to have one command start both servers as convenience mthod. That doesn't change anything else, or am I missing something?

If we really want to make a one command to run everything I think we'll just go with a `.sh` file which orchestrate everything.

sdmg15:
Inside `BmpServiceImpl::initialize` there is still hardcoded testenv, what would be the best approach to make it production ready? Should we pass the chain params through the proto request?

rpc/src/test/java/bisq/BmpServiceIntegrationTest.java
```diff
@@ -21,19 +23,51 @@ public class BmpServiceIntegrationTest {
...
+        // Mine some blocks to ensure test environment has funds
+        System.out.println("Mining initial blocks...");
+        testenv.mineBlocks(101);
+    }
+
+    /**
+     * Wait for a gRPC service to be ready by attempting to call a method
+     */
+    private void waitForGrpcReady(BmpProtocolServiceGrpc.BmpProtocolServiceBlockingStub stub, int maxAttempts) {
+        for (int i = 0; i < maxAttempts; i++) {
```
ChrisSon15, line 66: This is not really calling a gRPC method???

rpc/src/test/java/bisq/BmpServiceIntegrationTest.java
```diff
@@ -21,19 +23,51 @@ public class BmpServiceIntegrationTest {
...
+            // Try to ping by calling getBalance with a dummy request
+            System.out.println("  Checking gRPC server readiness (attempt " + (i + 1) + "/" + maxAttempts + ")");
+            Uninterruptibles.sleepUninterruptibly(500, TimeUnit.MILLISECONDS);
+            if (i >= 2) return;
+            if (i < maxAttempts - 1) {
+                Uninterruptibles.sleepUninterruptibly(500, TimeUnit.MILLISECONDS);
+            }
+        }
+        System.out.println("  gRPC server appears ready (or timeout reached)");
```
ChrisSon15: timeout and 'ready' should not be the same message.

rpc/src/test/java/bisq/BmpServiceIntegrationTest.java
```diff
@@ -178,12 +212,23 @@ void testHappyPathFullProtocolAndBroadcast() throws InterruptedException {
...
+
         // Give the network a moment to process the transaction, then mine a block.
         Thread.sleep(1000);
-        nigiri.mineBlocks(1);
+        testenv.mineBlocks(1);
         System.out.println("Mined a block to confirm the transaction.");
 
-        String rawTx = nigiri.getRawTransaction(depositTxid);
+        // Wait for transaction to appear in RPC
+        if (!testenv.waitForTransaction(depositTxid)) {
```
ChrisSon15, line 241: Since this is a testenvironmet, shouldnt we make this throw an exception?!

testenv/src/bin/testenv-server.rs
```diff
@@ -0,0 +1,79 @@
...
+    let electrum_url = env.electrum_url();
+    let workdir = env.workdir();
+
+    // Output as key=value pairs for easy parsing
+    println!("TESTENV_READY=true");
+    println!("TESTENV_RPC_URL={}", rpc_url);
+    println!("TESTENV_RPC_PORT={}", rpc_port);
+    println!("TESTENV_RPC_USER=bitcoin");
+    println!("TESTENV_RPC_PASS={}", env.bitcoin_rpc_password());
+    println!("TESTENV_ELECTRUM_URL={}", electrum_url);
```
ChrisSon15, line 42: we shouldnt export the electrum URL at this point. Its probably shouldnt be used in future

ChrisSon15:
> Created this to quickly gather feedback. The new way of running the rpc will be:
> 
> 1- Run the bitcoind client:
> 
> `cargo run --bin testenv-server -p testenv -- --data-dir "$DATA_DIR"`
> 
> 2- From the output you'll be able to get the RPC node connection information to be used next
> 
> ```shell
> cargo run --bin musigd --manifest-path rpc/Cargo.toml -- \
>   --port 50051 \
>   --bitcoin-rpc-url=http://127.0.0.1:37459 \
>   --bitcoin-rpc-user=bitcoin \
>   --bitcoin-rpc-pass=passs
> ```
> 
> 3- `mvn clean verify `
> 
> But there is an issue: inside `MemWallet::funded_wallet` on line 33 we're launching another testenv in order to fund the wallet. Why not just reusing the same wallet that's created inside `musigd.rs` main function (created while instantiating `WalletImpl`) that way we would be able to fund it inside `musigd.rs`

I guess we need to rewrite the readme.md in rpc as well, right?

rpc/src/bin/musigd.rs
```diff
@@ -9,29 +10,83 @@ use rpc::pb::bmp_protocol::bmp_protocol_service_server::BmpProtocolServiceServer
...
+
+    /// Bitcoin Core RPC password
+    /// Can also be set via BITCOIN_RPC_PASS environment variable.
+    #[arg(long)]
+    bitcoin_rpc_pass: Option<String>,
+
+    /// Electrum server URL (optional, for wallet sync optimization)
+    /// Can also be set via ELECTRUM_URL environment variable.
+    #[arg(long)]
+    electrum_url: Option<String>,
```
ChrisSon15, line 42: electrum_url is never used, or did I overlook something?

ChrisSon15 (changes requested):
Also the readme.md has stale content with these changes.

ChrisSon15:
> Inside `BmpServiceImpl::initialize` there is still hardcoded testenv, what would be the best approach to make it production ready? Should we pass the chain params through the proto request?

Good point. I am not sure if this bmp_service wont end up as dead code. Really is a question for @stejbac 
Lets move bmp_service.rs to test section, it has not only TestEnv, but also references to MemWallet and Testchain. I am not sure its worth investing time into it, because it may never be productive code.

rpc/src/bin/musigd.rs
sdmg15: Yes for now it's not, its usage is now depending on how we decide to handle `BmpServiceImpl::initialize` function. Right now the `initialize` function is launching another testenv instance although at time of its execution there's already a running testenv instance.

testenv/src/lib.rs
```diff
@@ -97,9 +97,61 @@ impl Default for Config<'_> {
...
+            std::fs::create_dir_all(data_dir).map_err(|e| {
+                anyhow::anyhow!(
+                    "Failed to create data directory {}: {}",
+                    data_dir.display(),
+                    e
+                )
+            })?;
+
+            unsafe {
+                std::env::set_var("ELECTRSD_DIR", data_dir);
```
stejbac: I don't think we should be using any unsafe code at all without a code comment to explain why it is sound.

According to the Rustdoc for `set_var`, it's unsafe to use in a multi-threaded program on Unix, since concurrent reads of the environment from other threads (via libc or the Rust standard library) cannot be practically prevented.

rpc/src/test/java/bisq/TestEnvClient.java
```diff
@@ -0,0 +1,325 @@
...
+     */
+    public double getBalance() {
+        String result = rpcCall(
+                "getbalance",
+                "[]"
+        );
+        try {
+            return Double.parseDouble(extractResult(result));
+        } catch (Exception e) {
+            throw e;
```
stejbac: Why are we catching an exception only to re-throw it? The try-catch block would appear to be redundant.

stejbac:
> Good point. I am not sure if this bmp_service wont end up as dead code. Really is a question for @stejbac Lets move bmp_service.rs to test section, it has not only TestEnv, but also references to MemWallet and Testchain. I am not sure its worth investing time into it, because it may never be productive code.

Yes, I think it may be best to move `bmp_service.rs` to the test section and remove the `BmpProtocolServiceServer` instance from the configured list of gRPC services in `musigd.rs`, since it will likely never be used in production and it seems like it's too much work to break its dependency on `TestEnv`, `MemWallet` and `Testchain`. However, I don't think the whole of `musigd` should be moved to the test section, since I always intended it to be a production binary, along with `musig-cli`. They should probably be moved into their own package(s) at some point, though, instead of making `rpc` a mixed library & binary package as it is presently -- `rpc` should probably be library-only, like the other packages.

ChrisSon15:
> > Good point. I am not sure if this bmp_service wont end up as dead code. Really is a question for @stejbac Lets move bmp_service.rs to test section, it has not only TestEnv, but also references to MemWallet and Testchain. I am not sure its worth investing time into it, because it may never be productive code.
> 
> Yes, I think it may be best to move `bmp_service.rs` to the test section and remove the `BmpProtocolServiceServer` instance from the configured list of gRPC services in `musigd.rs`, since it will likely never be used in production and it seems like it's too much work to break its dependency on `TestEnv`, `MemWallet` and `Testchain`. However, I don't think the whole of `musigd` should be moved to the test section, since I always intended it to be a production binary, along with `musig-cli`. They should probably be moved into their own package(s) at some point, though, instead of making `rpc` a mixed library & binary package as it is presently -- `rpc` should probably be library-only, like the other packages.

I agree. Musigd should be refactored and moved back to production code. But not in this PR. It should be done in a followup PR. We are trying to get sdmg's PR through and in that context this PR is a sub-PR to sdmgs PR. So we are trying to make minimal changes. All real refactoring work on RPC, I would like to see in a next PR (made by whom?), so we can quickly get this PR through. The move on musigd to test section is a no-brainer for me because it has a reference the TestEnv, and therefore is test code, unless we start rewriting it, which as said is not in the scope of this PR.

stejbac:
> I agree. Musigd should be refactored and moved back to production code. But not in this PR. It should be done in a followup PR. We are trying to get sdmg's PR through and in that context this PR is a sub-PR to sdmgs PR. So we are trying to make minimal changes. All real refactoring work on RPC, I would like to see in a next PR (made by whom?), so we can quickly get this PR through. The move on musigd to test section is a no-brainer for me because it has a reference the TestEnv, and therefore is test code, unless we start rewriting it, which as said is not in the scope of this PR.

I'm a little confused. This PR clashes with PR #160 to move `musigd` to the test section, since this PR adds a new `testenv-server` binary target and adds some extra command-line parameters to `musigd` to pick up from the TestEnv server output, whereas PR #160 removes the `musigd` binary altogether, moving its logic into the test section. So in what sense is PR #160 a followup to this PR? And what exactly will be the role of the TestEnv server if both this PR and PR #160 are merged?

ChrisSon15:
> I'm a little confused. This PR clashes with PR #160 to move `musigd` to the test section, since this PR adds a new `testenv-server` binary target and adds some extra command-line parameters to `musigd` to pick up from the TestEnv server output, whereas PR #160 removes the `musigd` binary altogether, moving its logic into the test section. So in what sense is PR #160 a followup to this PR? And what exactly will be the role of the TestEnv server if both this PR and PR #160 are merged?

Ops, I was confused. Actually I worked off the wrong branch, I should have used this PR as base line for moving bmp_service.rs to Test section. Sorry for messing up. So lets ignore PR#160.

ChrisSon15:
> 
> Ops, I was confused. Actually I worked off the wrong branch, I should have used this PR as base line for moving bmp_service.rs to Test section. Sorry for messing up. So lets ignore PR#160.

@sdmg15 can you move the bmp_service.rs to test and remove the reference in musigd?
@stejbac could you decide on the future of the bmp-classes?

rpc/src/bin/musigd.rs
```diff
@@ -1,47 +1,101 @@
...
+    }
+    if cli.bitcoin_rpc_user.is_none() {
+        cli.bitcoin_rpc_user = std::env::var("BITCOIN_RPC_USER").ok();
+    }
+    if cli.bitcoin_rpc_pass.is_none() {
+        cli.bitcoin_rpc_pass = std::env::var("BITCOIN_RPC_PASS").ok();
+    }
+    if cli.electrum_url.is_none() {
+        cli.electrum_url = std::env::var("ELECTRUM_URL").ok();
+    }
```
stejbac: I don't think we should mirror each new command-line argument with an environment variable, as that adds needless complexity to the code (and I guess that's not really a standard practice). I think we should be keeping the use of environment variables to a minimum, as least for the production code.

rpc/src/bin/musigd.rs
```diff
@@ -1,47 +1,101 @@
...
+        {
+            Auth::UserPass(user.clone(), pass.clone())
+        } else {
+            // Try cookie file as fallback
+            Auth::CookieFile(std::path::PathBuf::from("~/.bitcoin/.cookie"))
+        };
+
+        BitcoinCoreClient::new(rpc_url, auth)?
+    } else {
+        return Err("Can't proceed without bitcoin_rpc_url".into())
```
stejbac, line 64: Since `WalletServer` isn't being used for anything important and will probably never be used in production, I don't think we should make `bitcoin_rpc_url` a mandatory argument, as that makes it harder to run `musigd`. It should either fall back to the default Regtest RPC URL of `http://127.0.0.1:18443` or exclude `WalletServer` from the configured list of gRPC services, if it's missing.

If a default is provided for the `bitcoin_rpc_url` command-line argument, that should probably be specified in the Clap attributes, so that it shows up in the help (and because it's simpler).

rpc/Cargo.toml
```diff
@@ -28,11 +28,9 @@ unimock = { version = "0.6.8", optional = true }
 # Dependencies used only by the binary target(s):
 # TODO: Consider making a workspace of separate packages to avoid pulling these into the library:
 bmp_tracing = { workspace = true }
-clap = { version = "4.6.0", features = ["derive"] }
 wallet = { path = "../wallet" }
-# TODO move this to dev-dependencies — `bmp_service.rs` currently uses `Testchain` as a
-# placeholder `ChainApi` until `rpc` switches to a real chain backend.
-testenv = { path = "../testenv" }
+clap = { version = "4.6.0", features = ["derive"] }
```
stejbac, line 31: Why was this line moved down. (Also a spurious newline has been added.)

rpc/pom.xml
```diff
@@ -86,7 +103,7 @@
                 <artifactId>protobuf-maven-plugin</artifactId>
                 <version>0.6.1</version>
                 <configuration>
-                    <protocArtifact>com.google.protobuf:protoc:3.25.6:exe:${os.detected.classifier}</protocArtifact>
+                    <protocArtifact>com.google.protobuf:protoc:3.25.5:exe:${os.detected.classifier}</protocArtifact>
```
stejbac: Why was this downgraded?

rpc/pom.xml
```diff
@@ -45,6 +46,22 @@
...
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
+            <version>3.25.5</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java-util</artifactId>
+            <version>3.25.5</version>
+        </dependency>
```
stejbac, line 53: I'm wondering if we really need well-known types support if `google/protobuf/empty.proto` has been removed from the proto imports. I think that was the only use of well-known types.

testenv/src/lib.rs
```diff
@@ -97,9 +97,59 @@ impl Default for Config<'_> {
...
+            // Create the data directory if it doesn't exist
+            std::fs::create_dir_all(data_dir).map_err(|e| {
+                anyhow::anyhow!(
+                    "Failed to create data directory {}: {}",
+                    data_dir.display(),
+                    e
+                )
+            })?;
+
+            unsafe { std::env::set_var("ELECTRSD_DIR", data_dir) };
```
stejbac: I don't think we should be using any unsafe code at all without a code comment to explain why it is sound.

According to the Rustdoc for `set_var`, it's unsafe to use in a multi-threaded program on Unix, since concurrent reads of the environment from other threads (via libc or the Rust standard library) cannot be practically prevented.

testenv/src/lib.rs
```diff
@@ -97,9 +97,59 @@ impl Default for Config<'_> {
...
+        }
+
+        TestEnv::new_with_conf(self.config)
+    }
+}
+
+impl Default for TestEnvBuilder {
```
stejbac: Wouldn't it be simpler to derive `Default` and make `new()` delegate to `default()`, rather than the other way round?

rpc/src/bin/musigd.rs
```diff
@@ -1,47 +1,101 @@
...
-use testenv::TestEnv;
 use tonic::transport::Server;
 
 #[derive(Debug, Parser)]
 #[command(version, about, long_about = None)]
-#[expect(clippy::doc_markdown, reason = "doc comments are used verbatim by Clap and not intended to be markdown")]
+#[expect(
+    clippy::doc_markdown,
+    reason = "doc comments are used verbatim by Clap and not intended to be markdown"
+)]
```
stejbac, line 18: Could you avoid making spurious reformatting of unchanged lines if possible? There are a few minor incompatibilities between rustfmt and my IDE autoformatter, even though the latter is fairly lax, and it doesn't like this particular multi-line style.

rpc/README.md
```diff
@@ -78,46 +78,23 @@ mvn exec:java -Pwallet
...
 ```
 
-as usual, which will automatically start Nigiri (with the `--ci` flag to exclude Esplora) if it isn't running already.
-
-It continues to run after the tests finish, so to stop Nigiri, subsequently run:
-
-```sh
-nigiri --datadir "$PWD/.nigiri" stop
-```
+Take not of the Bitcoinc RPC server credentials as they'll be used for starting the
```
stejbac: "Take note"?

stejbac (changes requested):
I added some in-code comments/suggestions.

Since this PR appears to be combined with some of the changes in PR #160, I'm a little confused about the exact role of the TestEnv server. I take it that it's needed to run one of the Java integration tests (`BmpServiceIntegrationTest`)? But don't the individual Alice and Bob gRPC server instances start their own TestEnv instance each (as they're run as `cargo test -p rpc --test musigd ...`)? So does that mean that there are three TestEnv instances running?

sdmg15:
> I added some in-code comments/suggestions.
> 
> Since this PR appears to be combined with some of the changes in PR #160, I'm a little confused about the exact role of the TestEnv server. I take it that it's needed to run one of the Java integration tests (`BmpServiceIntegrationTest`)? But don't the individual Alice and Bob gRPC server instances start their own TestEnv instance each (as they're run as `cargo test -p rpc --test musigd ...`)? So does that mean that there are three TestEnv instances running?

Yes that's true at the end we have three instances running but that's just because of the `initialize`, else each instances of the Alice/Bob is using the same testenv already run by `cargo run --bin testenv-server -p testenv -- --data-dir "$DATA_DIR"`

testenv/src/lib.rs
sdmg15: I'm not sure about the comment I need to put there? As you said the Rustdoc already have the description, or should I put that description as comment?

PR #158, opened 2026-06-05
refactor protocol to use generics instead of trait objects for ProtocolWalletApi and TradeWallet

ChrisSon15:
introduced a sync_all method in ProtocolWalletApi to initiate a wallet sync from protocol trade. This was only possible with changing the TradeWallets to be generics instead of trait object. Since Rust prefers generics and there is no real reason to have 2 types of Wallet objects at the same runttime, I converted the 'impl TradeWallet' to generics. The changes look like a lot, but are mainly formatting changes. The real changes are pretty straight forward.

To keep this simple, I changed funded_wallet to return only one type. in more detail:

refactor(protocol): make trade protocol generic over wallet type

  Replace `Box<dyn TradeWallet>` with a static type parameter `W: TradeWallet`
  so the protocol is statically dispatched over its wallet. This is required
  because `ProtocolWalletApi::sync_all<D: ChainDataSource>` is a generic method,
  which makes the trait not dyn-compatible.

  - BMPContext/BMPProtocol become BMPContext<W>/BMPProtocol<W>; thread <W> through the per-tx helper methods. Drop the BoxedTradeWallet alias.
  - Unify mock_buyer/seller_trade_wallet to a single concrete type (MockWallet, backed by Vec iterators) so a generic TradeModel is possible.
  - rpc TradeModel<W = MockWallet> with a default type param; pin bmp_service to BMPProtocol<MemWallet>.
  - Pin integration tests to the BMPWallet backend (drop WALLET_BACKEND switch).

### DIFF
--- a/protocol/src/protocol_musig_adaptor.rs
+++ b/protocol/src/protocol_musig_adaptor.rs
@@ -10,7 +10,7 @@ use musig2::secp::{MaybeScalar, Point};
 use musig2::{PartialSignature, PubNonce};
 
 use crate::multisig::{KeyCtx, PointExt as _, SigCtx};
-use crate::psbt::BoxedTradeWallet;
+use crate::psbt::TradeWallet;
 use crate::receiver::{Receiver, ReceiverList};
 use crate::script_paths;
 use crate::transaction::{
@@ -87,19 +87,20 @@ pub struct Round4Parameter {
 }
 
 /// this context is for the whole process and need to be persisted by the caller
-pub struct BMPContext {
+pub struct BMPContext<W: TradeWallet> {
     // first of all, everything which is general to the protocol itself.
-    // `funds` is type-erased so the protocol works against any [`TradeWallet`]
-    // implementation (e.g. `MemWallet`, `BMPWallet`, mocks).
-    pub funds: BoxedTradeWallet,
+    // `funds` is the concrete wallet the protocol operates against (e.g. `MemWallet`,
+    // `BMPWallet`, mocks); it is a generic type parameter rather than a trait object so the
+    // protocol stays statically dispatched over the chosen [`TradeWallet`] implementation.
+    pub funds: W,
     pub chain: Box<dyn ChainApi>,
     pub role: ProtocolRole,
     pub seller_amount: Amount,
     pub buyer_amount: Amount,
 }
 
-pub struct BMPProtocol {
-    pub ctx: BMPContext,
+pub struct BMPProtocol<W: TradeWallet> {
+    pub ctx: BMPContext<W>,
     // Point securing Seller deposit and trade amount:
     pub p_tik: KeyCtx,
     // Point securing Buyer deposit:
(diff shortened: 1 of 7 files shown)


stejbac:
I think it's a bad idea to use generics for `TradeWallet` instead of trait objects, and that this PR is too disruptive to code that I am actively working on. We should favour dynamism and late binding while the protocol code is immature, for maximum flexibility. The generics are contagious and ever more type parameters are likely to be added as more functionality is implemented, making the code complex and hard to maintain.

I would greatly prefer and appreciate much more conservative changes being made to the protocol side while I'm actively working on it, keeping `TradeWallet` dyn-compatible if at all possible. That would require `ChainDataSource` to be made dyn-compatible if a `sync_all` method is to be added, which I understand could be a little tricky, since it has associated constants and `BMPWalletPersister` has an associated type parameter. But maybe it should be dyn-compatible anyway, since the chain backend (Electrum, CBFs, RPC, etc.) is decided at runtime and might even switch during runtime.

It was always my intention to keep the `TradeWallet` API fairly minimal. Does a `sync_all` method even belong as part of that API? `TradeWallet` should function as a black box to generate addresses & internal keys for the protocol, build the Deposit PSBTs, sign PSBTs and import private keys at the end of the trade, which are all the protocol's concerns. It's not clear to me that keeping the wallet synced with the chain is really the protocol's concern.

I haven't worked out the best way to handle chain notification, broadcast and syncing, with the former two as an abstract API to be consumed by the protocol, but I'm of the strong opinion that they should all be asynchronous, and I'm undecided whether it should be part of the `TradeWallet` API or an entirely separate API and trait object.

ChrisSon15:
> I think it's a bad idea to use generics for `TradeWallet` instead of trait objects, and that this PR is too disruptive to code that I am actively working on. We should favour dynamism and late binding while the protocol code is immature, for maximum flexibility. The generics are contagious and ever more type parameters are likely to be added as more functionality is implemented, making the code complex and hard to maintain.
> 
> I would greatly prefer and appreciate much more conservative changes being made to the protocol side while I'm actively working on it, keeping `TradeWallet` dyn-compatible if at all possible. That would require `ChainDataSource` to be made dyn-compatible if a `sync_all` method is to be added, which I understand could be a little tricky, since it has associated constants and `BMPWalletPersister` has an associated type parameter. But maybe it should be dyn-compatible anyway, since the chain backend (Electrum, CBFs, RPC, etc.) is decided at runtime and might even switch during runtime.
> 
> It was always my intention to keep the `TradeWallet` API fairly minimal. Does a `sync_all` method even belong as part of that API? `TradeWallet` should function as a black box to generate addresses & internal keys for the protocol, build the Deposit PSBTs, sign PSBTs and import private keys at the end of the trade, which are all the protocol's concerns. It's not clear to me that keeping the wallet synced with the chain is really the protocol's concern.
> 
> I haven't worked out the best way to handle chain notification, broadcast and syncing, with the former two as an abstract API to be consumed by the protocol, but I'm of the strong opinion that they should all be asynchronous, and I'm undecided whether it should be part of the `TradeWallet` API or an entirely separate API and trait object.

I wasn't aware that these changes have such an impact on your code you are currently writing. So yes, we can roll it back and not have `sync_all` in the Api. At some point we will probably want the blockchain events drive the sync-updates anyway.

As far as the theoretical discussion trait object vs generics goes, I dont see why we would need more than one chain backend at any time. For production there will be probably one and that wont be switchable. Only for testing we could need some some testing backends. So on a requirement level I dont see yet why we would need dyn Traits.

If you want to see where I was coming from, I added the function to ProtocolWalletApi:
`fn sync_all(&mut self, data_source: &impl ChainScanner)` and received a compile error saying: 
``TradeWallet` is not dyn compatible` any more.

PR #160, opened 2026-06-10
refactor(rpc): move musigd daemon and bmp_service into integration test

ChrisSon15:
Remove the rpc/src/bin/musigd.rs binary and the rpc::bmp_service library
  module, relocating both into rpc/tests/musigd.rs so the full gRPC server
  (and its BMP trade-protocol service impl) can be started as a test case.

  - test_musigd_starts_and_serves: smoke test on an OS-assigned port
  - run_musigd_server: #[ignore]d long-running server on MUSIGD_PORT (default 50051), replacing `cargo run --bin musigd -- --port <PORT>`
  - move BmpServiceImpl alongside the daemon (its only consumer)
  - testenv is now test-only -> moved to dev-dependencies
  - update rpc/README.md to use the new `cargo test` invocations


stejbac:
`musigd` was always intended to be used in production, when I first created it. It's currently being used for manual testing against Bisq2, for the testing and development of the Bisq2 trade protocol code against the `Musig` gRPC service, but later hopefully a wallet gRPC service as well. Bisq2 MuSig development requires a _production_ gRPC Rust server to test against.

Instead of moving the entire daemon to the test folder, I think we should just strip out everything from the `rpc` crate except the `musigrpc.Musig` service and the BMP `wallet.Wallet` service, which don't currently depend on TestEnv. Everything else (`bmp_protocol.BmpProtocolService` and `walletrpc.Wallet`) is basically legacy code at this point and will likely never be used.

sdmg15 (approved):

ChrisSon15:
> `musigd` was always intended to be used in production, when I first created it. It's currently being used for manual testing against Bisq2, for the testing and development of the Bisq2 trade protocol code against the `Musig` gRPC service, but later hopefully a wallet gRPC service as well. Bisq2 MuSig development requires a _production_ gRPC Rust server to test against.
> 

The functionality of musigd still exists, only the syntax calling musigd has changed. This has been made to get the test functionality into the test section. And the code right now only uses testservers. If you think this should be further refactured, I am all with you and think the usage of a server should be a parameter so we can use it againt a productive or against a test server. But I think that should be a second step and the PR should be merged and as it improves the code. Any additional features should be address in a followup PR IMHO.

> Instead of moving the entire daemon to the test folder, I think we should just strip out everything from the `rpc` crate except the `musigrpc.Musig` service and the BMP `wallet.Wallet` service, which don't currently depend on TestEnv. Everything else (`bmp_protocol.BmpProtocolService` and `walletrpc.Wallet`) is basically legacy code at this point and will likely never be used.

Sorry, I dont find the structs you are mentioning. It looks they are gRPC generated classes, which doesnt make sense. May be you can elaborate on this.

ChrisSon15:
got refactorred and redone in PR#155

PR #161, opened 2026-06-23
(feat)using testenv in rpc

ChrisSon15:
follow-up of PR#155, added some findings regarding the usage of testenv with musigd CLI.


sdmg15 (approved):
tAck
I have a follow up PR for #155 too, so once this is merged I'm opening it.

PR #162, opened 2026-06-24
fix: failling rpc tests 

sdmg15:
This decouples the `bmp_service.rs` test from Testenv. The wallet is funded from the available RPC commands.
Previously `bmp_service.rs` when run was launching another bitcoind which wasn't connected with the other ones making the tests to fail.

This PR also allow defaulting password and some parameters to ease the running commands.


testenv/src/lib.rs
```diff
@@ -130,7 +132,7 @@ impl TestEnvBuilder {
                 )
             })?;
         }
-
+        self.config.password = Some("bitcoin".to_string());
```
ChrisSon15: If you want to set a default password, shoudnt we do so in theTestEnvBuilder::new() method?

rpc/tests/bmp_service.rs
```diff
@@ -40,10 +38,75 @@ use tonic::{Request, Response, Result, Status, transport};
...
+    electrum_url: Option<String>,
+}
+
+struct BitcoindChain {
+    client: Arc<BitcoinCoreClient>,
+}
+
+impl ChainApi for BitcoindChain {
+    fn transaction_broadcast(&self, tx: &Transaction) -> anyhow::Result<Txid> {
+        self.client.send_raw_transaction(tx).map_err(Into::into)
```
ChrisSon15: In the Testchain broadcast_via() method there is an explicit softening of retransmitting the same transaction twice. Why are you removing it?
Its been in there because the 2 traders can broadcast the DepositTx both independently into different blocks. Current testing is not testing this edge case.

rpc/tests/bmp_service.rs
```diff
@@ -40,10 +38,75 @@ use tonic::{Request, Response, Result, Status, transport};
...
 
-#[derive(Default)]
 pub struct BmpServiceImpl {
     // Each trade protocol is stored against a unique ID.
     active_protocols: Mutex<HashMap<String, BMPProtocol>>,
+    bitcoin_rpc_client: Arc<BitcoinCoreClient>,
+    electrum_url: Option<String>,
+}
+
+struct BitcoindChain {
```
ChrisSon15: Testchain is actually indendently usable from TestEnv. As far as I can see, you dont need BitcoindChain, just reuse Testchain. Or are you trying to get rid of the testEnv dependency?

rpc/tests/bmp_service.rs
```diff
@@ -40,10 +38,75 @@ use tonic::{Request, Response, Result, Status, transport};
...
+        "Wallet failed to sync funded balance after {MAX_RETRIES} attempts"
+    ))
+}
+
+impl BmpServiceImpl {
+    pub fn new(client: Arc<BitcoinCoreClient>, electrum_url: Option<String>) -> Self {
+        Self {
+            active_protocols: Mutex::new(HashMap::new()),
+            bitcoin_rpc_client: client,
+            electrum_url,
```
ChrisSon15, line 95: Why is electrum_url Optional, I think it cant work if None??

ChrisSon15 (commented):
i uess the findings in bmp_service are not so important as its unclear where we are going with it.

sdmg15:
> i uess the findings in bmp_service are not so important as its unclear where we are going with it.

The goal was to fix the failling rpc issues actually. So at this state it's working well, and we won't have to deal with the bmp_service.rs anymore

PR #167, opened 2026-07-06
Update dependencies and cleanup

stejbac:
Make formatting changes to assorted Rust and Java files, to make the latter a little more idiomatic (and compliant with IDEA's autoformatter) and to make the former a little more consistent (at least per file) as well bringing some of the IDE autoformatting slightly closer to rustfmt compatibility (namely spacing around `=`).

Also make some other shallow code cleanup, in particular fixing some mostly-pedantic-level Clippy warnings in the `testenv` and `wallet` packages, so that all the project packages build cleanly with the workspace-configured lints, which are now enabled for all packages.

Additionally, run `cargo update` to bring all the direct dependencies to latest except `electrsd` & `bdk_electrum` (which seem to be difficult to make a major update to without inadvertently enabling AWS-LC in place of Ring for `rustls`), `rusqlite`, `bdk_kyoto` & `bdk_wallet` (which would move BDK to version 3 and presumably be a big change), and `hmac` & `sha2` (for which a major update would prevent reuse of the lagging `hmac` & `sha2` dependencies of `musig2`/`secp`).

In particular `musig2` and `secp` are updated to 0.4 and 0.7 respectively, which is a breaking change requiring some minor changes to the `protocol::multisig` module to incorporate the secret key in the MuSig2 nonce seed. Updating Rust Bitcoin required `FeeRate::from_sat_per_vb_unchecked` to be replaced to prevent deprecation warnings. And `bdk_wallet` is updated to 2.4.0 instead of being pinned to 2.1, now that `bdk_wallet::wallet::signer` has been un-deprecated.

Finally, fix a deadlock in the `testenv` tests when they are run single-threaded, due to the penultimate `TestEnv` instance in `test_persistent_data_dir` needing to be dropped before creating a final `TestEnv` instance.

### DIFF
--- a/protocol/src/multisig.rs
+++ b/protocol/src/multisig.rs
@@ -48,9 +48,9 @@ struct NoncePair {
 }
 
 impl NoncePair {
-    fn new(nonce_seed: impl Into<NonceSeed>, aggregated_pub_key: Point) -> Self {
-        let sec_nonce = SecNonceBuilder::new(nonce_seed)
-            .with_aggregated_pubkey(aggregated_pub_key)
+    fn new(nonce_seed: impl Into<NonceSeed>, seckey: Scalar, aggregated_pubkey: Point) -> Self {
+        let sec_nonce = SecNonceBuilder::from_seckey(nonce_seed, seckey)
+            .with_aggregated_pubkey(aggregated_pubkey)
             .build();
         Self { pub_nonce: sec_nonce.public_nonce(), sec_nonce: Some(sec_nonce) }
     }
@@ -220,10 +220,11 @@ impl SigCtx {
     }
 
     pub fn init_my_nonce_share(&mut self) -> Result<()> {
-        let aggregated_pub_key = self.tweaked_key_ctx()?.key_agg_ctx.aggregated_pubkey();
+        let seckey = self.tweaked_key_ctx()?.my_prv_key;
+        let aggregated_pubkey = self.tweaked_key_ctx()?.key_agg_ctx.aggregated_pubkey();
         // TODO: Consider making the RNG configurable, to aid unit testing:
         self.my_nonce_pair_share.get_or_insert_with(||
-            NoncePair::new(&mut rand::rng(), aggregated_pub_key));
+            NoncePair::new(&mut rand::rng(), seckey, aggregated_pubkey));
         Ok(())
     }
 
--- a/rpc/src/test/java/bisq/TestEnvClient.java
+++ b/rpc/src/test/java/bisq/TestEnvClient.java
@@ -14,11 +14,13 @@ import java.util.regex.Pattern;
 
 /**
  * Bitcoin RPC client for TestEnv integration tests
- * <p>
- * Environment variables:
- * - TESTENV_RPC_URL
- * - TESTENV_RPC_USER
- * - TESTENV_RPC_PASS
+ *
+ * <p>Environment variables:
+ * <ul>
+ *   <li>{@code TESTENV_RPC_URL}
+ *   <li>{@code TESTENV_RPC_USER}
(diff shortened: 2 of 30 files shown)


sdmg15:
> bdk_kyoto & bdk_wallet (which would move BDK to version 3 and presumably be a big 

This is done in PR#168

protocol/src/multisig.rs
```diff
@@ -220,10 +220,11 @@ impl SigCtx {
     }
 
     pub fn init_my_nonce_share(&mut self) -> Result<()> {
-        let aggregated_pub_key = self.tweaked_key_ctx()?.key_agg_ctx.aggregated_pubkey();
+        let seckey = self.tweaked_key_ctx()?.my_prv_key;
+        let aggregated_pubkey = self.tweaked_key_ctx()?.key_agg_ctx.aggregated_pubkey();
         // TODO: Consider making the RNG configurable, to aid unit testing:
         self.my_nonce_pair_share.get_or_insert_with(||
-            NoncePair::new(&mut rand::rng(), aggregated_pub_key));
+            NoncePair::new(&mut rand::rng(), seckey, aggregated_pubkey));
```
sdmg15, line 227: Is this now taking three params due to version upgrades?

rpc/src/test/java/bisq/TestEnvClient.java
```diff
@@ -13,14 +14,13 @@
 
 /**
  * Bitcoin RPC client for TestEnv integration tests
- *
+ * <p>
```
sdmg15: Pardon my ignorance here, what's this p for? If it's opened shouldn't it be closed too?

stejbac: Blank lines are ignored in Javadoc because it's HTML, and `<p>` doesn't need to be closed because HTML is less strict about closing tags than XML. Most Javadoc I've seen in libraries doesn't seem to bother with paragraph closing tags.

protocol/src/multisig.rs
stejbac: `NoncePair::new` uses `musig2::SecNonceBuilder` to build a secret nonce. But with the version upgrade, the secret nonce now holds a copy of the local pubkey share as an extra field (for better BIP 327 compliance), so either the local pubkey or or the corresponding secret key must be passed to the builder and are no longer optional seed material for the nonce. (The aggregated pubkey is still optional seed material, but I decided to keep it in to make the nonce generation slightly more robust.)

rpc/src/test/java/bisq/TestEnvClient.java
stejbac: Since the original doc-comment appears to be markdown, which unfortunately isn't supported before Java 23 (with `///` lines), perhaps I should have reformatted it as something like:
```
/**
 * Bitcoin RPC client for TestEnv integration tests
 *
 * <p>Environment variables:
 * <ul>
 *   <li>{@code TESTENV_RPC_URL}
 *   <li>{@code TESTENV_RPC_USER}
 *   <li>{@code TESTENV_RPC_PASS}
 * </ul>
 */
```

stejbac:
I've pushed another commit to fix broken Testchain transaction broadcast idempotency, whereby the Electrum error that a tx has already been seen on the blockchain is no longer being swallowed, due to the error message changing with Bitcoin Core 28.

sdmg15 (approved):
ACK 22287a2a7e0f4de1b03323fe628b332dd448031a

stejbac:
I've also pushed a commit to change the `TestEnvClient` Javadoc from Markdown to HTML, per the earlier comment, and rewritten the fix in the last commit slightly differently.

ChrisSon15:
utACK, as I am traveling

PR #168, opened 2026-07-07
Refactor compact block filter syncing. Unify

sdmg15:
This PR also introduces several changes
- bump bdk_wallet to 3
- bump bdk_kyoto to 0.16
- make the production code completely free from testenv

### DIFF
--- a/chain/Cargo.toml
+++ b/chain/Cargo.toml
@@ -6,6 +6,10 @@ edition = "2024"
 [dependencies]
 anyhow = { workspace = true }
 bdk_wallet = { workspace = true }
+tokio = { workspace = true }
+bmp_tracing = { workspace = true }
+bdk_kyoto = { workspace = true }
+
 
 [lints]
 workspace = true
--- a/chain/Cargo.toml
+++ b/chain/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2024"
 anyhow = { workspace = true }
 bdk_wallet = { workspace = true }
 tokio = { workspace = true }
-bmp_tracing = { workspace = true }
+tracing = { workspace = true }
 bdk_kyoto = { workspace = true }
 
 
--- a/chain/src/lib.rs
+++ b/chain/src/lib.rs
@@ -1,8 +1,10 @@
 use std::collections::BTreeMap;
 use std::sync::Arc;
 
-use bdk_kyoto::bip157::Network;
-use bdk_kyoto::{BuilderExt as _, Info, Receiver, ScanType, UnboundedReceiver, Update, Warning};
+use bdk_kyoto::bip157::{Builder, Network};
+use bdk_kyoto::{
+    BuilderExt as _, Info, Receiver, ScanType, TrustedPeer, UnboundedReceiver, Update, Warning,
+};
 use bdk_wallet::Wallet;
 use bdk_wallet::bitcoin::{Address, Amount, Transaction, Txid};
 use bdk_wallet::chain::DescriptorId;
@@ -37,15 +39,15 @@ pub trait ChainScanner {
     ) -> anyhow::Result<FullScanResponse<K>>;
 }
 
-pub struct CBFScanner;
-
-impl Default for CBFScanner {
-    fn default() -> Self {
-        Self
-    }
+pub struct CBFScanner {
+    pub peers: Vec<TrustedPeer>,
 }
 
 impl CBFScanner {
+    pub const fn new(peers: Vec<TrustedPeer>) -> Self {
+        Self { peers }
+    }
+
     async fn traces(
         mut info_subscriber: Receiver<Info>,
         mut warning_subscriber: UnboundedReceiver<Warning>,
@@ -77,13 +79,17 @@ impl CBFScanner {
     pub async fn sync_cbf(
         &self,
         network: Network,
+        peers: Vec<TrustedPeer>,
         wallets: Vec<(&Wallet, ScanType)>,
     ) -> anyhow::Result<BTreeMap<DescriptorId, Update>> {
-        let client = bdk_kyoto::bip157::Builder::new(network).build_with_wallets(wallets)?;
+        let client = Builder::new(network)
+            .add_peers(peers)
+            .build_with_wallets(wallets)?;
 
         let (client, logging, mut update_subscriber) = client.subscribe();
+
         tokio::task::spawn(async move {
-            Self::traces(logging.info_subscriber, logging.warning_subscriber).await
+            Self::traces(logging.info_subscriber, logging.warning_subscriber).await;
         });
         let client = client.start();
         let requester = client.requester();
--- a/chain/src/lib.rs
+++ b/chain/src/lib.rs
@@ -1,12 +1,21 @@
+use std::collections::BTreeMap;
 use std::sync::Arc;
 
-use bdk_wallet::bitcoin::{Transaction, Txid};
+use bdk_kyoto::bip157::Network;
+use bdk_kyoto::{BuilderExt as _, Info, Receiver, ScanType, UnboundedReceiver, Update, Warning};
+use bdk_wallet::Wallet;
+use bdk_wallet::bitcoin::{Address, Amount, Transaction, Txid};
+use bdk_wallet::chain::DescriptorId;
 use bdk_wallet::chain::spk_client::{FullScanRequest, FullScanResponse};
+use bmp_tracing::tracing;
+use tokio::select;
 
 /// Abstraction over blockchain interaction for broadcasting transactions.
 /// to be extended
 pub trait ChainApi: Send + Sync {
     fn transaction_broadcast(&self, tx: &Transaction) -> anyhow::Result<Txid>;
+    fn send_to_address(&self, address: &Address, amount: Amount) -> anyhow::Result<()>;
+    fn generate_to_address(&self, blocks: u32, address: &Address) -> anyhow::Result<()>;
 }
 
 /// Abstraction over the read-side of a chain backend: pre-populating a transaction cache and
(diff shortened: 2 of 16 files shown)


testenv/Cargo.toml
```diff
@@ -9,6 +9,7 @@ anyhow = { workspace = true }
 bdk_bitcoind_rpc = { workspace = true }
 bdk_electrum = { workspace = true }
 bdk_wallet = { workspace = true, features = ["test-utils"] }
+wallet = { path = "../wallet"}
```
stejbac, line 12: This creates a circular dependency between packages, if not at least between the underlying crates. This is presumably to provide the `Testchain` impl of `ChainDataSource`, which must live in either `testenv` or `wallet`, without having a production dependency of `wallet` on `testenv`.

An alternative might be to make `testenv` an optional production dependency of `wallet` and feature-gate the `Testchain` impl, but maybe this approach is the simplest for packages in the same workspace (and `testenv` is already dependent on `chain`). I'm not sure what the better approach would be. @ChrisSon15, what do you think?

sdmg15: I don't think that's worth the trouble. `testenv` is already in `dev-deps` of `wallet` and `wallet` is in main `dependencies` of `testenv`.

chain/Cargo.toml
```diff
@@ -6,6 +6,10 @@ edition = "2024"
 [dependencies]
 anyhow = { workspace = true }
 bdk_wallet = { workspace = true }
+tokio = { workspace = true }
+bmp_tracing = { workspace = true }
```
stejbac: I think library crates should only import the `tracing` facade, rather than `bmp_tracing`, since a tracing subscriber isn't needed and otherwise becomes undesirable indirect dependency of the library.

protocol/Cargo.toml
```diff
@@ -11,14 +11,17 @@ musig2 = { workspace = true }
 paste = "1.0.15"
 rand = { workspace = true }
 rand_chacha = { workspace = true }
+
```
stejbac: A spurious newline has been added here.

testenv/Cargo.toml
stejbac: Yes, that's what I was saying -- this PR introduces a circular (package) dependency, as `wallet` was not in the main dependencies of `testenv` before these changes.

I was suggesting moving the `Testchain` impl of `ChainDataSource` from `testenv` back to `wallet` and feature-gating it, so that the subsequent production dependency of `wallet` on `testenv` is only optional. Then presumably there would be no need to make `testenv` dependent on `wallet`. (But as that's slightly more complicated, it might not obviously be better.)

wallet/src/protocol_wallet_api.rs
```diff
@@ -1,12 +1,15 @@
...
+use std::time::Duration;
 
 use bdk_electrum::BdkElectrumClient;
 use bdk_electrum::bdk_core::bitcoin::bip32::Xpriv;
-use bdk_electrum::bdk_core::bitcoin::{XOnlyPublicKey, absolute, secp256k1};
+use bdk_electrum::bdk_core::bitcoin::{
+    Address, Amount, FeeRate, Network, OutPoint, Psbt, ScriptBuf, XOnlyPublicKey,
+};
 use bdk_electrum::electrum_client::Client;
-use bdk_wallet::bitcoin::{Address, Amount, FeeRate, Network, OutPoint, Psbt, ScriptBuf};
```
stejbac, line 9: What is the reason for importing these types via `bdk_electrum` rather than `bdk_wallet`?

(In fact, the existing `use bdk_electrum::bdk_core::bitcoin::bip32::Xpriv` import should probably be made via `bdk_wallet` as well, since it is a more fundamental dependency than `bdk_electrum` and `Xpriv` isn't being used for anything Electrum related.)

wallet/tests/wallet_integration_test.rs
```diff
@@ -53,15 +54,15 @@ fn test_sync_with_imported_keys() -> anyhow::Result<()> {
...
 
-    wallet.sync_all(&chain)?;
+    wallet.sync_all(&chain).await?;
     assert_eq!(wallet.balance(), receive_amount + receive_amount);
 
     Ok(())
 }
 
-#[test]
+#[tokio::test]
```
stejbac: Not part of these changes, but there's a spurious newline here.

wallet/tests/wallet_integration_test.rs
```diff
@@ -231,11 +232,11 @@ async fn test_cbf_main_wallet() -> anyhow::Result<()> {
 
     env.mine_blocks(4)?;
 
-    let scan_type = bdk_kyoto::ScanType::Sync;
-    let peers = vec![TrustedPeer::from_socket_addr(
+    let _scan_type = bdk_kyoto::ScanType::Sync;
```
stejbac: Why is `_scan_type` being brought into scope? The enum `ScanType` has no significant drop, so there's no need for an RAII guard here.

wallet/tests/wallet_integration_test.rs
```diff
@@ -259,11 +260,12 @@ async fn test_cbf_imported() -> anyhow::Result<()> {
 
     env.mine_blocks(4)?;
 
-    let scan_type = bdk_kyoto::ScanType::Sync;
+    let _ = bdk_kyoto::ScanType::Sync;
```
stejbac: Same here -- `ScanType` doesn't serve as an RAII guard (and would be immediately dropped here anyway).

wallet/tests/wallet_integration_test.rs
```diff
@@ -290,13 +292,12 @@ async fn test_cbf_imported_and_main() -> anyhow::Result<()> {
...
```
stejbac: Same here.

wallet/tests/wallet_integration_test.rs
```diff
@@ -315,12 +316,14 @@ async fn test_cbf_persistence() -> anyhow::Result<()> {
     let addr = wallet.next_unused_address(KeychainKind::External);
     env.fund_address(&addr, Amount::from_sat(230_000))?;
 
-    let scan_type = bdk_kyoto::ScanType::Sync;
-    let peers = vec![TrustedPeer::from_socket_addr(
+    let _scan_type = bdk_kyoto::ScanType::Sync;
```
stejbac: Same here.

wallet/Cargo.toml
```diff
@@ -17,14 +17,11 @@ rusqlite = { workspace = true }
...
+thiserror =  { workspace = true }
+trait-variant = { workspace =  true }
 
 [dev-dependencies]
 bdk_wallet = { workspace = true, features = ["test-utils"] }
 bmp_tracing = { workspace = true }
 tempfile = { workspace = true }
-
-[lints]
-workspace = true
```
stejbac, line 30: Was it intended to move `wallet` back to default rather than workspace lints. I think all the crates should be using the pedantic workspace lints. (It would have caught two of the above-commented `ScanType` code defects, for example.)

stejbac (changes requested):
There are a number of shallow issues I picked up, though I didn't spot any higher level problems, and it's desirable to be making more of the IO-performing operations async.

wallet/tests/wallet_integration_test.rs
```diff
@@ -259,11 +258,12 @@ async fn test_cbf_imported() -> anyhow::Result<()> {
 
     env.mine_blocks(4)?;
 
-    let scan_type = bdk_kyoto::ScanType::Sync;
+    let _ = bdk_kyoto::ScanType::Sync;
```
stejbac: I'm disappointed that Clippy still isn't picking this up. The [no_effect_underscore_binding](https://rust-lang.github.io/rust-clippy/master/index.html#no_effect_underscore_binding) lint apparently doesn't apply to non-binding wildcards, only underscore-prefixed variables.

This line should be removed, as it does nothing -- the evaluation of the RHS is side-effect free and the drop is trivial.

wallet/tests/wallet_integration_test.rs
```diff
@@ -290,13 +290,12 @@ async fn test_cbf_imported_and_main() -> anyhow::Result<()> {
...
```
stejbac: This still needs to be addressed, as it is redundant code not being picked up by Clippy for some reason (as above).

wallet/src/chain_data_source.rs
```diff
@@ -52,7 +52,7 @@ impl ChainDataSource for CBFScanner {
 
         for (descriptor, update) in updates {
             let idx = *descriptors_map.get(&descriptor).unwrap();
-            let w = &mut *wallets.get_mut(idx).unwrap();
+            let w = &mut **wallets.get_mut(idx).unwrap();
```
ChrisSon15: this is not code for testing, so we should start to ban unwrap() and similar panics.

chain/src/lib.rs
```diff
@@ -1,12 +1,22 @@
...
+use bdk_wallet::bitcoin::{Address, Amount, Transaction, Txid};
+use bdk_wallet::chain::DescriptorId;
 use bdk_wallet::chain::spk_client::{FullScanRequest, FullScanResponse};
+use tokio::select;
 
 /// Abstraction over blockchain interaction for broadcasting transactions.
 /// to be extended
 pub trait ChainApi: Send + Sync {
     fn transaction_broadcast(&self, tx: &Transaction) -> anyhow::Result<Txid>;
+    fn send_to_address(&self, address: &Address, amount: Amount) -> anyhow::Result<()>;
```
ChrisSon15, line 19: these are 2 fn for testing only, they should not be part of the official API

chain/src/lib.rs
```diff
@@ -27,3 +37,84 @@ pub trait ChainScanner {
...
+            .updates()
+            .await?
+            .collect::<BTreeMap<_, _>>();
+
+        requester.shutdown()?;
+        Ok(updates)
+    }
+}
+
+impl ChainScanner for CBFScanner {
```
ChrisSon15: why would we need these unimplemented methods?

wallet/src/chain_data_source.rs
```diff
@@ -1,36 +1,60 @@
...
+    ) -> anyhow::Result<()>;
 }
 
-/// `ChainDataSource` implementation for the Electrum-backed [`testenv::Testchain`] handle.
-///
-/// `Testchain` lives in the `testenv` crate alongside `TestEnv`; this impl ties it into the
-/// wallet's sync routine. It mirrors the values previously used by the live Electrum-backed
-/// implementation.
-impl ChainDataSource for testenv::Testchain {
+impl ChainDataSource for CBFScanner {
```
ChrisSon15, line 22: these consts are probably not needed for the CBFScanner?!

wallet/src/bmp_wallet.rs
```diff
@@ -388,26 +356,85 @@ pub trait WalletApi {
...
+        }
+
+        self.imported_balance = final_imported_balance;
+
+        // Persist changes from imported keys
+        for (w, db) in &mut imported {
+            w.persist(db)?;
+        }
+
+        let _ = self.persist();
```
ChrisSon15: The Result of self.persist is silently swallowed, is a '?' missing?

chain/src/lib.rs
```diff
@@ -27,3 +37,84 @@ pub trait ChainScanner {
...
+    }
+
+    async fn traces(
+        mut info_subscriber: Receiver<Info>,
+        mut warning_subscriber: UnboundedReceiver<Warning>,
+    ) {
+        loop {
+            select! {
+                info = info_subscriber.recv() => {
+                    if let Some(info) = info {
```
ChrisSon15, line 58: This is missing the else case. After channel close recv() still returns None here and the loop needs to break.

ChrisSon15 (commented):
Soory it took some time to do the review, please look at the comment for API changes

wallet/src/chain_data_source.rs
sdmg15: Yes not needed but should be defined.

wallet/src/chain_data_source.rs
```diff
@@ -1,36 +1,66 @@
...
 
-    fn sync(&self, persister: &mut PersistedWallet<impl BMPWalletPersister>) -> anyhow::Result<()> {
-        let tx_nodes = persister.tx_graph().full_txs().map(|tx_node| tx_node.tx);
-        self.populate_tx_cache(tx_nodes);
-        let request = persister.start_full_scan();
+    async fn sync(
+        &self,
+        mut wallets: Vec<&mut PersistedWallet<impl BMPWalletPersister>>,
+    ) -> anyhow::Result<()> {
+        let network = wallets[0].network();
```
ChrisSon15, line 31: could panic

ChrisSon15 (approved):
I will merge as is, generally looks good.

In the subsequent work we should address:
wallet::bmp_wallet::BMPWallet::import_private_key  should error when same key tried to import twice

Some test code is still in production section, we should move it and have the separation better.

wallet/src/chain_data_source.rs
sdmg15: Well nope, it's guaranteed not too, as there's always a MAIN WALLET.

stejbac: Since it's technically a public API, the potential panic if incorrectly passed an empty Vec should probably be documented as part of the contract of the `ChainDataSource::sync` trait method, in the Panics section of its Rustdoc.

I disabled the corresponding `missing_panics_doc` pendatic-level Clippy lint, since it flags too many false positives, being unable to distinguish cases where panics are part of the function contract (that is, it's a partial function, panicking on some inputs, as here) from cases where no panic actually occurs no matter what arguments are passed to the function.

PR #169, opened 2026-07-12
Remove `TradeWallet` trait and move `MockTradeWallet` to its own module

stejbac:
Implement the stubbed-out `ProtocolWalletAPI::create_pbst` method for `MockTradeWallet`, which the `TradeWallet::create_half_deposit_psbt` method goes through by default, instead of directly overriding it. In this way, `TradeWallet` becomes a (redundant) pure extension trait and can be removed, replacing it's `create_half_deposit_psbt` extension method with a crate-visible utility fn, used only by `DepositTxBuilder`. Replace all uses of `TradeWallet` in the code with `ProtocolWalletAPI`, using `impl/dyn ProtocolWalletAPI` in the interfaces in place of `impl/dyn TradeWallet`.

Also move `MockTradeWallet` and its impls into their own `protocol::mocks` module (for use by some of the unit tests and the `rpc` crate), so that `protocol::psbt` becomes a private utility module (for the tx builders in `protocol::transaction`).

Additionally, fix some of the error handling in `protocol::psbt`, and make `ProtocolWalletAPI` use `Result<T, WalletErrorKind>` for its fallible methods in place of `anyhow`. Also deduplicate the code in the `MemWallet` impl of `ProtocolWalletAPI` slightly, by making it a pure delegate to (the `ProtocolWalletAPI` interface of) its internal BDK wallet.

### DIFF
--- a/protocol/src/mocks.rs
+++ b/protocol/src/mocks.rs
@@ -1,0 +1,222 @@
+use std::collections::BTreeMap;
+
+use bdk_wallet::bitcoin::taproot::Signature;
+use bdk_wallet::bitcoin::transaction::Version;
+use bdk_wallet::bitcoin::{
+    Address, Amount, FeeRate, Network, OutPoint, Psbt, ScriptBuf, Sequence, TapSighashType,
+    Transaction, TxIn, TxOut, VarInt, Weight, Witness, XOnlyPublicKey, absolute, psbt,
+};
+use musig2::secp::Scalar;
+use wallet::protocol_wallet_api::{ProtocolWalletApi, WalletErrorKind};
+
+use crate::mocks::WalletErrorKind::Other;
+use crate::psbt::Redact as _;
+use crate::transaction::{TransactionErrorKind, TxOutput};
+
+struct MockTradeWallet<Cs: Iterator<Item = TxOutput>, As: Iterator<Item = Address>> {
+    funding_coins: Cs,
+    new_addresses: As,
+    signature_map: BTreeMap<OutPoint, Signature>,
+    internal_key: Option<XOnlyPublicKey>,
+    script_sigs: BTreeMap<XOnlyPublicKey, Vec<Signature>>,
+}
+
+impl<Cs: Iterator<Item = TxOutput>, As: Iterator<Item = Address>> ProtocolWalletApi for MockTradeWallet<Cs, As> {
+    fn network(&self) -> Network { Network::Regtest }
+
+    fn new_address(&mut self) -> Result<Address, WalletErrorKind> {
+        self.new_addresses.next().ok_or_else(|| Other(TransactionErrorKind::MissingAddress.into()))
+    }
+
+    fn new_internal_key(&mut self) -> Result<XOnlyPublicKey, WalletErrorKind> {
+        self.internal_key.take().ok_or_else(|| Other(TransactionErrorKind::MissingAddress.into()))
+    }
+
+    fn create_psbt(
+        &mut self,
+        mut recipients: Vec<(ScriptBuf, Amount)>,
+        fee_rate: FeeRate,
+    ) -> Result<Psbt, WalletErrorKind> {
+        let fee_cost_msat = |weight: Weight|
+            fee_rate.to_sat_per_kwu().checked_mul(weight.to_wu())
+                .ok_or(Other(TransactionErrorKind::Overflow.into()));
+
+        // Provisionally add a change recipient of zero value. We should never normally use
+        // `new_address()` for change outputs, but this is just a mock.
+        recipients.push((self.new_address()?.script_pubkey(), Amount::ZERO));
+
+        let base_weight = Weight::from_wu_usize(38 + 4 * VarInt::from(recipients.len()).size());
+        let mut output = Vec::with_capacity(recipients.len());
+        let mut cost_msat = fee_cost_msat(base_weight)?;
+
+        for (script_pubkey, value) in recipients {
+            let tx_out = TxOut { value, script_pubkey };
+            cost_msat = (|| cost_msat
+                .checked_add(value.to_sat().checked_mul(1000)?)?
+                .checked_add(fee_cost_msat(tx_out.weight()).ok()?))()
+                .ok_or(Other(TransactionErrorKind::Overflow.into()))?;
+            output.push(tx_out);
+        }
+
+        let mut input = Vec::new();
+        let mut inputs = Vec::new();
+        let mut funds = Amount::ZERO;
+
+        while funds < Amount::from_sat(cost_msat.div_ceil(1000)) {
+            let new_coin = self.funding_coins.next()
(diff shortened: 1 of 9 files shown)


protocol/src/psbt.rs
```diff
@@ -406,25 +149,26 @@ fn half_psbt_fee_overpay_msat(psbt: &Psbt, target_fee_rate: FeeRate) -> Option<i
...
+        signed_tx_weight += coin.estimated_input_weight()
+            .ok_or(TransactionErrorKind::InvalidPsbt)?;
+        input_amount = input_amount.checked_add(coin.prevout.value)
+            .ok_or(TransactionErrorKind::Overflow)?;
     }
-    let output_amount = psbt.unsigned_tx.output.iter().map(|o| o.value).checked_sum()?;
-
-    let actual_fee_msat = input_amount.checked_sub(output_amount)?.to_sat().checked_mul(1000)?;
-    let target_fee_msat = target_fee_rate.to_sat_per_kwu().checked_mul(signed_tx_weight.to_wu())?;
+    (|| {
```
ChrisSon15, line 165: You removed the FIXME from the old file, but still an overflow error is  reported instead of an invalid PSBT. Was there a special reason to delete the FIXME or just by accident?

wallet/src/protocol_wallet_api.rs
```diff
@@ -189,49 +183,38 @@ impl WalletExt for MemWallet {
...
+    ) -> Result<()> {
+        self.wallet.sign_selected_inputs(psbt, is_selected)
     }
 
-    fn import_private_key(&mut self, _pk: Scalar) {
-        // `MemWallet` is an in-memory wallet that doesn't currently support imported keys.
-        // If/when this is needed, mirror the `BMPWallet` implementation.
-        todo!("MemWallet does not yet support importing private keys")
+    fn import_private_key(&mut self, pk: Scalar) {
+        self.wallet.import_private_key(pk);
```
ChrisSon15: now we are getting an unimplemented! panic at this point. Since we have no intentions to implement the fn import_private_key for bdk_wallet, I dont understand how this change improved anything, to me the old todo! was more meaningful.

wallet/src/protocol_wallet_api.rs
```diff
@@ -133,10 +126,11 @@ impl MemWallet {
         self.wallet.balance().trusted_spendable()
     }
 
+    pub fn reveal_next_address(&mut self) -> AddressInfo {
+        self.wallet.reveal_next_address(KeychainKind::External)
+    }
+
     pub fn next_unused_address(&mut self) -> AddressInfo {
```
ChrisSon15, line 133: I dont understand why the FIXME is deleted, it seems the problems still exists.

protocol/src/mocks.rs
```diff
@@ -0,0 +1,222 @@
...
+use bdk_wallet::bitcoin::taproot::Signature;
+use bdk_wallet::bitcoin::transaction::Version;
+use bdk_wallet::bitcoin::{
+    Address, Amount, FeeRate, Network, OutPoint, Psbt, ScriptBuf, Sequence, TapSighashType,
+    Transaction, TxIn, TxOut, VarInt, Weight, Witness, XOnlyPublicKey, absolute, psbt,
+};
+use musig2::secp::Scalar;
+use wallet::protocol_wallet_api::{ProtocolWalletApi, WalletErrorKind};
+
+use crate::mocks::WalletErrorKind::Other;
```
ChrisSon15, line 12: why not import this as `use wallet::protocol_wallet_api::WalletErrorKind::Other;` ?

protocol/src/psbt.rs
stejbac: I intended overflows in the checked arithmetic to still be reported as `TransactionErrorKind::Overflow`, as they are elsewhere, but other errors such as invalid input coin types (not P2TR or P2WPKH) or too many tx inputs or outputs are reported as `TransactionErrorKind::InvalidPsbt`, instead of returning an `Option<i64>` so that every failure is reported as a blanket `OverflowError` (which was incorrect). Reporting a blanket `InvalidPsbt` error instead, say, would be inconsistent with the way overflows are handled elsewhere.

The errors are still very course-grained but I guess that would be better addressed in a separate PR.

wallet/src/protocol_wallet_api.rs
stejbac: The `bdk_wallet::Wallet::next_unused_address` method, which this same-named method delegates to, does what it was indented to. It's just its use in our code, in particular the protocol code to get new addresses for the fee bumps, claim payouts, etc., that was wrong. But that's going through `ProtocolWalletAPI::new_address` now, so the original FIXME is no longer applicable.

(I think its reasonable for the user to assume that `Memwallet::next_unused_address` has the same behaviour as `bdk_wallet::Wallet::next_unused_address`.)

protocol/src/mocks.rs
stejbac: I guess I could have done, but `WalletErrorKind` is already imported and `use crate::mocks::WalletErrorKind::Other;` is a bit shorter.

I wanted to import it just as `use WalletErrorKind::Other;`, but then rustfmt wants to canonicalise the imports by putting it near the top of the list, right under the `use std::...` block of imports, which seemed a bit out of place.

wallet/src/protocol_wallet_api.rs
stejbac: Yes, I think on reflection the original `todo!` should have been kept, to distinguish between something that won't ever be implemented and something that needs to be implemented. (Plus IDEs will flag `todo!` but not `unimplemented!`.)

I was probably a bit too eager to make the `ProtocolWalletAPI` impl of `Memwallet` an exact delegate to its inner wallet, instead of being willing to break the symmetry where appropriate.

ChrisSon15:
ah, one more thing, mocks.rs seems to be test code only. Shouldn't that go into the test area? We have lately been trying to move all test code into the proper section. Is that intended for next PR?

stejbac:
> ah, one more thing, mocks.rs seems to be test code only. Shouldn't that go into the test area? We have lately been trying to move all test code into the proper section. Is that intended for next PR?

The mock buyer/seller trade wallet is still used by the `musigd` binary for the `Musig` gRPC service, via `rpc::protocol`, which is used for testing/development against Bisq2 (and by the `TradeProtocolClient.java` test/demo). I guess once a project-local Regtest deployment via `make`/`just` or similar is in place, then `musigd` could be changed to use a real trade wallet. (Though actual protocol tx broadcasts and listening for them on chain still hasn't been implemented for the `Musigd` service, and I think I would like to develop an abstract interface for that, possibly driven by mocks, and try to use it in `rpc::protocol` before switching to a real backend.)

It may still be worth ultimately keeping mocks and fakes in the production code anyway, but behind a feature gate for use by the tests of downstream crates, like the feature-gated `bdk_wallet::test_utils` module.

PR #170, opened 2026-07-23
Add `TradeState` enum to improve protocol safety

stejbac:
Provide an enum of 16 distinct trade states, including 9 closure states, to mark each time that critical trade data needs persisting and prevent operations from being performed at the wrong time. The graph of legal moves between states forms a DAG, with edges defined by the relation, `TradeState::precedes(self, Self) -> bool`. (There are also buyer/seller-specific analogues `TradeState::{precedes_for_buyer, precedes_for_seller}`.) This is with the caveat that in the presence of tx replacement or block reorgs, the trade could get into conflicting states, which will likely need to be represented with a separate `TradeErrorState` struct/enum or similar (not yet implemented), and then illegal jumps between states might be possible. It's intended, however, that the trade never reopens once moved to one of the `TradeClosed(_)` states.

Update the protocol code in `rpc::protocol::TradeModel` to track the trade state, as an extra struct field, along with a private `TradeModel::update_state(&mut self, new_state: TradeState)` method to update and mock-persist the state with the rest of the trade data, returning an error for illegal state transitions. (Actual persistence is not implemented for the trade protocol yet.) The intention is to provide `TradeModel` with a completely safe API, regardless of the order its methods are called in, with trade-state-gated getters for sensitive data like the signed Deposit PSBT.

Also add unit tests that the DAG of `TradeState` transitions matches what's expected, and that its `PartialOrd` impl agrees with the partial order induced by the DAG (or at least the DAG's transitive reduction).

Finally, add a rendering `trade_state_transitions.png` of the graph of legal moves, generated using Graphviz from the test data: an embedded DOT file of the expected full DAG.

--

- ~~**TODO:** Improve the error handling of the `TradeModel` getters, instead of just returning `Option<T>`. In particular, provide a new `ProtocolErrorKind::AccessDenied` variant, or similar.~~
- **TODO:** Maybe the `Deposit` trade state should be broken up into two or three separate states, to distinguish when the Deposit Tx signatures have been released (and it therefore might appear on-chain at any time) from its actual appearance and/or confirmation on chain. I'm not totally sure about this, because it's really the Java client's responsibility not to allow the buyer to start payment until the Deposit Tx has actually confirmed. But the Rust server nonetheless needs to notify the client when that point is reached, so incorporating that information into the (Rust-side) trade state might simplify things.


protocol/src/state.rs
```diff
@@ -0,0 +1,276 @@
...
+            TradeClosed(ClosureType::Custom) => &[CustomPayoutSigned],
+            TradeClosed(ClosureType::BuyersRedirect) => &[SellersWarning],
+            TradeClosed(ClosureType::SellersRedirect) => &[BuyersWarning],
+            TradeClosed(ClosureType::BuyersClaim) => &[BuyersWarning, COOP],
+            TradeClosed(ClosureType::SellersClaim) => &[SellersWarning, COOP],
+            TradeClosed(ClosureType::BuyersPenaltyTx | ClosureType::SellersPenaltyTx) => &[COOP],
+        }
+    }
+
+    const fn bits(self) -> u16 {
```
ChrisSon15, line 124: This looks like a transitive closure which could be easily calculated using the algorithm of Warshall. Given the small number of adolescences in the DAG it will compute in neglectable time. The tranient closure could be used directly as partial Order. The advantage would be to not have these constants in the code, because they are difficult to verify and when more states are added by other devs its error-prone.

rpc/src/main/java/bisq/TradeProtocolClient.java
```diff
@@ -316,9 +314,23 @@ private void doRestOfTrade(String buyerTradeId, String sellerTradeId, ClosureTyp
...
-        } else if (closureType == ClosureType.UNCOOPERATIVE) {
+        } else if (closureType == ClosureType.FORCED | closureType == ClosureType.FORCED_AFTER_MEDIATION) {
             // Seller attempts to send Message F to buyer, then waits...
 
+            if (closureType == ClosureType.FORCED_AFTER_MEDIATION) {
+                // Meanwhile, buyer accepts solution from mediator (fiat received, 50% seller security deposit penalty).
+
+                var buyersCustomPayoutPsbt = stub.signCustomPayoutTx(CustomPayoutPsbtRequest.newBuilder()
+                        .setTradeId(buyerTradeId)
+                        .setSellersPayoutAmountExcludingFee(15_000)
```
ChrisSon15, line 325: For constants we need to replace later on it would be good to have a TODO comment at lease, so we can find it later. A method stub would be another option.

rpc/src/protocol.rs
```diff
@@ -614,25 +635,27 @@ impl TradeModel {
...
+    }
 
     pub fn combine_deposit_psbts(&mut self, other: Psbt) -> Result<()> {
         self.deposit_tx.builder.combine_psbts(other)?;
         Ok(())
     }
 
     pub fn get_signed_deposit_tx(&self) -> Option<Transaction> {
-        self.deposit_tx.builder.signed_tx().ok()
+        self.deposit_tx.builder.signed_tx().ok().filter(|_| self.state >= TradeState::Deposit)
```
ChrisSon15: I like the idea of checking the states in the methods. Instead of returning a Option, shouldnt we prefer a result with error message?
May be calling a special fn at the beginning of each fn like 'checkState(....)' so its obvious for the reader which state(s) we are expecting? If you planned that anyway, then disregard my comment.

ChrisSon15 (commented):
just some quick notes

rpc/src/protocol.rs
stejbac: Yes, I'm about to push a commit which changes the `TradeState` getters to return `Result<T, ProtocolErrorKind>` instead of `Option<T>`. It adds a private `access!` macro to check the trade state against a pattern and return a generic "access denied" error in case of mismatch.

rpc/src/main/java/bisq/TradeProtocolClient.java
stejbac: Since it's test/demo code illustrating how to use the API, I think it's better to keep it as concrete as reasonably possible, instead of abstracting out the hard-coded constants used in the cases/examples.

protocol/src/state.rs
stejbac: The way that it is currently written, at least it is simple to verify that it manifestly defines a valid partial order, inherited by the embedding of `TradeState` into the boolean lattice `{0,1}^10`, since that depends only on the chosen 10-bit patterns being distinct.

The transitive _reduction_ of the partial order is being unit tested against the embedded DOT file `const TRADE_STATE_TRANSITION_GRAPH: &str`, in the test `trade_state_partial_order_has_expected_minimal_dag`, after doing some simple string replacements to strip out the transitive (dashed) edges and de-colourise the graph. That uniquely specifies the partial order, so any mismatch between it and the transitive closure of the embedded DAG should be picked up by the unit tests.

Similarly the `precedes`/`precedes_for_buyer`,`precedes_for_seller` relations defining the full DAG of legal state transitions (as opposed to its transitive reduction/closure) is being unit tested against the embedded DOT file, in the test `trade_state_has_expected_full_dag`.

So while it may be a little tricky to update the bit patterns for anyone wanting to modify the DAG, the unit tests should provide complete coverage for any mismatch between the DAG and either the `PartialOrd` impl (implementing its induced partial order) or the `TradeState::precedes*` methods (specifying its actual edges).

stejbac:
I have some pending changes to add prepared Penalty Tx creation (signed upon trade closure, and included as an extra field in the `CloseTradeResponse` gRPC message), as well as tightening access to a couple of the `TradeModel` getters to address TODOs in the code, and make the wallet-invoking methods in `TradeModel` idempotent. I'll probably leave that to another PR focusing on adding the Warning/Redirect/Claim/Penalty paths to `TradeModel` and the gRPC interface.

Similarly, the remaining TODO concerning _possible_ breakup of the `Deposit` trade state would probably be better addressed in a separate PR, when I've had a chance to think better about the broadcasting and chain listening logic.

protocol/src/state.rs
ChrisSon15: I see what you are saying. In practice, if someone wants to change the DAG, he will change the DOT file until it suits the needs. Then he would need to manually regenerate the bits() array, which is complicated and probably only the two of us understand how to do it. Instead of the bits array in its current form, letting the transitive closure been calculated and checked for DAG properties (acyclic) is much easier even to program even the first time. So I dont really see what the  advantage is to use the bits() array.

stejbac: The advantage is that the present code is fairly short (and efficient) and it's very easy to verify that it defines a valid `PartialOrd` impl. (It's very common to mess up the comparison functions of partial and total orders.) It would likely be considerably more lines of code to parse the DAG (or rewrite it as a list) and efficiently compute its transitive closure. (Plus, I find it mildly distasteful to have a heavyweight comparison function for a discrete set of 16 states, even if it computes in negligible time.)

I'm not sure we should be making it _too_ easy to change the DAG, since any changes to it are prone to have serious and non-obvious implications. Is it really such a hard puzzle to come up with a list a bit patterns which embed a small partially-ordered set into a boolean lattice? (This is analogous to defining a _total_ order on a set by embedding it into the integers -- that is, numbering its elements.)

protocol/src/state.rs
```diff
@@ -0,0 +1,276 @@
...
+    #[default]
+    Init,
+    /// We have signed the Deposit PSBT and are ready to share it. The peer may or may not already
+    /// have our signatures on it. The Deposit Tx may or may not have been broadcast or confirmed.
+    Deposit,
+    /// - (If buyer) we are ready to release our lock on the escrow by sharing our partial signature
+    ///   on the Swap Tx, which the seller may or may not have already received.
+    /// - (If seller) we have already received the buyer's partial signature on the Swap Tx and
+    ///   persisted it, releasing their lock on the escrow.
+    BuyerReadyToRelease,
```
ChrisSon15, line 18: Not quite sure what is meant here. (If buyer) The partial sig on SwapTx is already shared when the DepositTx is signed. So this state should be before state Deposit? That doesnt make sense either?

(if Seller) again sig on SwapTx is exchanged before DepsoitTx is signed. What is meant by 'releasing their lock on escrow'?

ChrisSon15 (approved):

protocol/src/state.rs
stejbac: In the implementation in `rpc::protocol`, the seller only gets the buyer's partial signature on the Swap Tx after the buyer confirms he has started payment (but well before the seller has had a chance to confirm), passing it to the Rust server in the (first) `SignSwapTx` gRPC call, at which point the seller aggregates to verify she has a valid Swap Tx signature. (As long as the aggregated adaptor signature is valid, the _seller_ is guaranteed that the final Swap Tx signature will be valid.)

This prevents the seller from being able to prematurely release the escrow before the buyer has started payment. If the buyer's partial signature on the Swap Tx is invalid, that just means the escrow will be stuck, and will have to be spent though a Warning or Custom Payout Tx. (There would hopefully also be proof of misbehaviour on the part of the buyer, to pass to the arbitrator.)

So both traders have a lock on the escrow, with the buyer releasing first, then the seller. (If one wanted to open the trade without a buyer lock on the escrow, one would move the trade state straight from `Init` to `BuyerReadyToRelease`, skipping past the `Deposit` state, or moving `Init -> Deposit -> BuyerReadyToRelease` in rapid succession, and consider `BuyerReadyToRelease` to be the starting state of the fully opened trade rather than `Deposit`.)

ChrisSon15: Ah, ok makes sense. I was still thinking the sig for SwapTx is exchanged while Init. But it can be exchanged later on as well.

PR #174, opened 2026-08-07
Add Penalty Tx signing and improve `TradeModel` API safety

stejbac:
Add static/prepared Penalty Tx construction and signing to `rpc::protocol::TradeModel`, reusing the Claim payout address and the prepared tx feerate agreed at the trade start. The Penalty Tx is signed upon cooperative trade closure (or if the buyer sees the seller's Swap Tx), that is, once the full aggregated private key for our payout is learned. Pass the signed Penalty Tx to the Java client in an added `CloseTradeResponse.penaltyTx` proto field (for optional watchtower use).

Make a few other changes to `rpc.proto` to address a couple of security-related TODOs in the `TradeModel` impl, improving the API safety by tightening access to the `signed_swap_tx` and `my_private_key_share_for_peer_output` getters. This involves moving the `swapTx` proto field one message later from `SwapTxSignatureResponse` (second call, when the seller decides to release the escrow) to `CloseTradeResponse` (when the seller actually force-closes the trade), and also making `CloseTradeResponse.peerOutputPrvKeyShare` optional, so that the buyer does _not_ reveal his key share in the case that the trade was force-closed by the seller. Even though these are technically breaking changes to the `Musig` API, it doesn't break the Bisq2 client since it hasn't implemented the force-closure path yet.

Also make the wallet-mutating methods of `TradeModel` idempotent to minimise potential unused address churn, in the same pattern as making its other methods idempotent where possible.

--

Add an indirect integration test for the Penalty path by implementing it in `protocol::protocol_musig_adapter` as well (and making an associated fix there to the use of `p_tik` vs `q_tik` for the Warning escrows), and adding a test case to `protocol_integration_tests`.

--

Finally, make some assorted minor changes, such as using somewhat less arbitrary choices of the bit patterns in `protocol::TradeState::bits` that are closer to the adjacency matrix of the partial order. (I could have made the bit patterns exactly equal to the adjacency matrix and that would have also worked, but they would be non-minimal length.) Also tidy the Cargo manifests a little and bump the dependencies.

--

- ~~**TODO:** Following discussion at the team meeting, it may be better to pass the chosen Penalty Tx feerate from the client to the Rust server at trade closure time, as it will be less out-of-date than the generic prepared tx feerate decided at the trade start, by the time the Penalty Tx actually has to be used. This should be a fairly easy change, I think.~~

### DIFF
--- a/bmp_tracing/Cargo.toml
+++ b/bmp_tracing/Cargo.toml
@@ -5,8 +5,8 @@ edition = "2024"
 
 [dependencies]
 tracing = { workspace = true }
-tracing-subscriber = { workspace = true }
 tracing-core = { workspace = true }
+tracing-subscriber = { workspace = true }
 
 [lints]
 workspace = true
--- a/rpc/src/main/proto/rpc.proto
+++ b/rpc/src/main/proto/rpc.proto
@@ -134,8 +134,8 @@ message SwapTxSignatureRequest {
 }
 
 message SwapTxSignatureResponse {
-  bytes swapTx = 1;
-  bytes peerOutputPrvKeyShare = 2;
+  //bytes swapTx = 1; // removed
+  optional bytes peerOutputPrvKeyShare = 2;
 }
 
 message CloseTradeRequest {
@@ -145,7 +145,8 @@ message CloseTradeRequest {
 }
 
 message CloseTradeResponse {
-  bytes peerOutputPrvKeyShare = 1;
+  optional bytes peerOutputPrvKeyShare = 1;
+  optional bytes swapTx = 2;
 }
 
 message CustomPayoutPsbtRequest {
--- a/rpc/src/main/proto/rpc.proto
+++ b/rpc/src/main/proto/rpc.proto
@@ -142,6 +142,7 @@ message CloseTradeRequest {
   string tradeId = 1;
   optional bytes myOutputPeersPrvKeyShare = 2;
   optional bytes swapTx = 3;
+  optional uint64 penaltyTxFeeRate = 4;
 }
 
 message CloseTradeResponse {
--- a/rpc/src/main/proto/rpc.proto
+++ b/rpc/src/main/proto/rpc.proto
@@ -147,6 +147,7 @@ message CloseTradeRequest {
 message CloseTradeResponse {
   optional bytes peerOutputPrvKeyShare = 1;
   optional bytes swapTx = 2;
+  optional bytes penaltyTx = 3;
 }
 
 message CustomPayoutPsbtRequest {
--- a/rpc/src/protocol.rs
+++ b/rpc/src/protocol.rs
@@ -12,8 +12,8 @@ use protocol::multisig::{KeyCtx, KeyPair, PointExt as _, SigCtx};
 use protocol::receiver::{Receiver, ReceiverList};
 use protocol::state::{ClosureType, TradeState};
 use protocol::transaction::{
-    CustomPayoutTxBuilder, DepositTxBuilder, ForwardingTxBuilder, NetworkParams as _,
-    RedirectTxBuilder, TransactionExt as _, WarningTxBuilder,
+    ClaimTxBuilder, CustomPayoutTxBuilder, DepositTxBuilder, NetworkParams as _, RedirectTxBuilder,
+    SwapTxBuilder, TransactionExt as _, WarningTxBuilder,
 };
 use protocol::{mocks, script_paths};
 use thiserror::Error;
@@ -87,7 +87,7 @@ struct DepositTx {
 
 #[derive(Default)]
 struct SwapTx {
-    builder: ForwardingTxBuilder,
+    builder: SwapTxBuilder,
     input_sighash: Option<TapSighash>,
     input_sig_ctx: SigCtx,
 }
@@ -107,7 +107,7 @@ struct RedirectTx {
 
 #[derive(Default)]
 struct ClaimTx {
-    builder: ForwardingTxBuilder,
+    builder: ClaimTxBuilder,
     input_sig_ctx: SigCtx,
 }
 
--- a/rpc/src/protocol.rs
+++ b/rpc/src/protocol.rs
@@ -567,7 +567,7 @@ impl TradeModel {
     }
 
     pub fn sign_swap_tx_input_partial(&mut self, sighash: TapSighash) -> Result<()> {
-        let sighash = self.swap_tx.input_sighash.insert(sighash);
+        let sighash = self.swap_tx.input_sighash.get_or_insert(sighash);
         self.swap_tx.input_sig_ctx.sign_partial(*sighash)?;
         Ok(())
     }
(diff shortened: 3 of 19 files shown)


rpc/src/protocol.rs
```diff
@@ -435,6 +457,8 @@ impl TradeModel {
             txs.claim.builder.compute_unsigned_tx()?;
             std::mem::swap(&mut txs, &mut peer_txs);
         }
+        self.penalty_tx.builder.set_input(peer_txs.warning.builder.escrow()?.clone());
```
ChrisSon15: the input is always the seller side. In line 450 Peer_txs is the seller side, then line 458 swapping gets executed exactly twice, so peer_txs is again the seller.
you might want to consider abandon the for-loop and instead calling a new method twice with the seller and buyer parameters reversed. I think that would  be more readable.

ChrisSon15 (commented):
even though in draft mode, just spotted the bug in penalty_tx

rpc/src/protocol.rs
stejbac: Yes, that's right. I had already spotted that and was about to fix it in as part of my final commit to allow the Penalty Tx feerate to be optionally set at trade closure time.

Certainly I could have picked better local variable names than `txs` and `peer_txs`, as that led me to introduce the bug.

stejbac:
I've just pushed another commit to add an optional `CloseTradeRequest.penaltyTxFeeRate` field to the `Musig` gRPC interface, overriding the default prepared tx feerate used for the Swap/Claim/Redirect/Warning Txs set up at the trade start.

rpc/src/protocol.rs
```diff
@@ -255,6 +263,11 @@ impl TradeModel {
         self.swap_tx.builder.set_fee_rate(fee_rate);
     }
 
+    pub fn set_penalty_tx_fee_rate(&mut self, fee_rate: Option<FeeRate>) -> Result<()> {
+        self.penalty_tx.builder.set_fee_rate(fee_rate.unwrap_or(self.prepared_tx_fee_rate()?));
```
ChrisSon15, line 267: if a fee_rate is supplied, it can happen that the prepared_tx_fee_rate fails and therefore the whole method fails (beside the obvious performance loss of the unnecessary call). Is that intended?

rpc/src/server.rs
```diff
@@ -196,16 +192,21 @@ impl musig_server::Musig for MusigImpl {
...
                 trade_model.set_peer_private_key_share_for_my_output(peer_prv_key_share)?;
                 trade_model.aggregate_private_keys_for_my_output()?;
+                trade_model.compute_signed_penalty_tx()?;
                 trade_model.close_trade(ClosureType::Cooperative)?;
             } else if let Some(swap_tx) = request.swap_tx.try_proto_into()? {
                 // Buyer supplies a signed swap tx to the Rust server, to close our trade. (Mainly for
                 // testing -- normally the tx would be picked up from the bitcoin network by the server.)
                 trade_model.recover_seller_private_key_share_for_buyer_output(&swap_tx)?;
                 trade_model.aggregate_private_keys_for_my_output()?;
+                trade_model.compute_signed_penalty_tx()?;
```
ChrisSon15, line 209: Why do you compute the PenaltyTx here?
The buyer wants to close with a SwapTx. If SwapTx is confirmedPenaltyTx. If SwapTx stays unconfirmed (in Mempool), then we probably wont get the signature but if, we should not use it.

bmp_tracing/Cargo.toml
```diff
@@ -5,8 +5,8 @@ edition = "2024"
 
 [dependencies]
 tracing = { workspace = true }
-tracing-subscriber = { workspace = true }
 tracing-core = { workspace = true }
+tracing-subscriber = { workspace = true }
```
sdmg15, line 9: Is this done manually or there's a rule in the formatting that's doing this?

rpc/src/main/proto/rpc.proto
```diff
@@ -134,8 +134,8 @@ message SwapTxSignatureRequest {
 }
 
 message SwapTxSignatureResponse {
-  bytes swapTx = 1;
-  bytes peerOutputPrvKeyShare = 2;
+  //bytes swapTx = 1; // removed
```
sdmg15, line 137: Maybe we can just get rid of this field if removed. 
The `peerOutputPrvKeyShare` shouldn't start now from 1 or that'd be breaking.

rpc/src/protocol.rs
stejbac: Yes, I did first try using `Option::unwrap_or_else`, but that fairly obviously doesn't work as the lambda is fallible, and there's no `unwrap_or_else_try` or similar method to use. So it would have made the code several lines longer, for not much benefit, as the prepared tx feerate is set early in the trade setup and `self.prepared_tx_fee_rate()` will always return `Ok(..)` by the time the penalty tx feerate is decided. The performance cost of eagerly evaluating the default (even when unused) is negligible.

rpc/src/server.rs
stejbac: Hypothetically, the Swap Tx could get replaced by the seller's Warning Tx (at least assuming we see unconfirmed txs, or there's a block reorg), in which case the buyer's Penalty Tx is still needed. Replacement is especially a concern for the buyer, as the Swap Tx is freeform from his point of view (since it's blinded and he only knows the sighash spending the seller's payout), so it could be paying any fee and have any additional inputs and could therefore get easily replaced.

For ergonomics, I don't think the buyer should wait for broadcast Swap Tx finality (which is just a line in the sand) before using its signature and marking the trade as closed.

stejbac:
> PenaltyTx and ClaimTx are so similar, that I get a lot of dulication warnings. A good way to get rid of the code duplication would be to have a shared base for both types, as they rely of ForwardingTxBuilder anyway. So only the differences in the timelock and the signing remain in the contrete types and the rest could go into a base struct, referenced as a field.

I take it you mean in the impl blocks of `protocol_musig_adaptor::{PenaltyTx, ClaimTx}`, where they have very similar `build`, `new`, `signed_tx` and `broadcast` methods (and the structs are structurally identical). I guess that could be deduplicated, though I'm not sure its really worth it, as I think it inappropriately couples the two txs and their internal structure together, and they should be free to diverge as they serve rather different functions.

I just tried deduplicating them by making them type aliases of a common `ClaimOrPenaltyTx<const T2: u16>` struct. It does save about 22 lines of code, but I'm really not sure if it's better code or if that's really what you had in mind.

(I don't think anything related can be usefully deduplicated in `rpc::protocol`.)

bmp_tracing/Cargo.toml
stejbac: I reformatted it manually, to try to keep the dependency lists in alphabetical order (except where there's a reason to order them differently).

rpc/src/main/proto/rpc.proto
stejbac: I thought it was better to leave it commented out until the Bisq2 client is updated, so that no one replaces its slot in the proto with a different incompatible field, in the meantime. One alternative would have been to mark the field number as reserved, but then the reservation reason might not be clear, so I couldn't really see the point. As soon as the client is brought up to sync with the Rust proto, the field number can be repurposed.

